PR #64, opened 2014-06-21
Convert fen param to string before split

felquis:
Hi, I'm using Chess.js and I think that this little fix will be appreciated

I suggest just to convert the fen parameter to a string before try to split it, to avoid errors when trying to split something that hasn't split method.

Current behavior: 
screen shot 2014-06-21 at 3 32 33 am

With this fix:
screen shot 2014-06-21 at 3 39 58 am

Best

### DIFF
--- a/chess.js
+++ b/chess.js
@@ -190,7 +190,7 @@ var Chess = function(fen) {
   }
 
   function load(fen) {
-    var tokens = fen.split(/\s+/);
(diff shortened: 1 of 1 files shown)


jhlywa:
Thanks for the pull request, but I agree with @gushogg-blake on this one.

felquis:
I agree, thanks :)

Issue #78, opened 2015-03-26
Pawn promotion without equal sign feature has been removed?


ngocdaothanh:
@davidmpaz
@jhlywa

There's this pull request:
https://github.com/jhlywa/chess.js/pull/33

But the latest code in master doesn't allow pawn promotion without equal sign any more. Has this feature been removed?

Now chess.js understands "fxg8=Q", but not "fxg8Q".

ngocdaothanh:
I've sent a PR. Please review.

jhlywa:
Can you add please add a quick test to the PR?  After all, that's how the "lazy parse" feature got dropped in the first place.

I eventually want to reevaluate how we load PGN in chess.js.  I'm of the opinion that ```load_pgn()``` should fail on non-standard SAN. Lazy parsing (or some better name) could be achieved by calling ```load_pgn({lazy: true})```

ngocdaothanh:
Thanks.
I'll add some test.

This is not only related to PGN loading.
This is also related to the `move` method.

ngocdaothanh:
I've added the test. Please review.

jhlywa:
Good catch on the ```move()``` method as well.  Thanks for your contribution

PR #79, opened 2015-03-26
Allow parsing pawn promotion without the equal sign (again)

ngocdaothanh:
Fix for https://github.com/jhlywa/chess.js/issues/78


jhlywa:
Thanks for the PR and tests

PR #81, opened 2015-04-04
Oriented prototype.

juangl:
I think that improves memory performance.


chinanzio:
Hello, i am developing a little chess application here http://standardchess.com
I found this API very helpful, and i changed some things i needed ... i would like to share it
I changed the code to make it use of prototype like follows

```
function Chess() {
}

Chess.prototype.someFunction = function() {
 // code here
};
```

i did it because the way it is, right now, creates new functions for every instance of Chess, which is not a problem if there is only one instance, but for servers, where there are, maybe thousands, it is better to have only one instance per function for every instance of Chess ....
Check out the following example. it demonstrates the difference

```
function Class_01(value) {
	this.value = value;
	this.someFunction = function() {
		console.log(this.value);
	};
}

function Class_02(value) {
	this.value = value;
}
Class_02.prototype.someFunction = function() {
	console.log(this.value);
};

var c01 = new Class_01(1);
var c02 = new Class_01(2);

var c03 = new Class_02(3);
var c04 = new Class_02(4);

console.log(c01.someFunction == c02.someFunction);
console.log(c03.someFunction == c04.someFunction);
```

the output will be:
false
true

Which means, that there is only one instance of someFunction for every instance of Chess
There is a second reason why i changed the code, "the static members". The correct way is to have them attached to the class definition rather than per Chess instance.
For example

```
function Chess() {
}

Chess.STATICMEMBER = 23;
```

I attach the modified source here, i don't know if this is the right place, guide me
[chess.js.zip](https://github.com/jhlywa/chess.js/files/273405/chess.js.zip)

If you want i can do it with the last version of chess.js and fix a difference in the Chess.turn member which i use directly, and the original chess.js uses as a function

`var chess = new Chess();`
My code uses `chess.turn` but the original code uses `chess.turn()`

And the last one, i added support for languages, because i speak spanish and the san notation changes from english, just to mention it

Take me to the right place if this shouldn't be here
Cheers

jhlywa:
Hi @chinanzio.  I'm in the process of converting to a prototype-based library.  It should be done soon and marks the transition to version 1.0.0.   I'm very interested in a translation layer for chess.js.  Do you have translations for the common PGN headers?

```
[Event "F/S Return Match"]
[Site "Belgrade, Serbia Yugoslavia|JUG"]
[Date "1992.11.04"]
[Round "29"]
[White "Fischer, Robert J."]
[Black "Spassky, Boris V."]
[Result "1/2-1/2"]
```

chinanzio:
Hi @jhlywa , do you mean spanish translation ? 

[Evento "F/S Return Match"]
[Sitio "Belgrade, Serbia Yugoslavia|JUG"]
[Fecha "1992.11.04"]
[Ronda "29"]
[Blancas "Fischer, Robert J."]
[Negras "Spassky, Boris V."]
[Resultado "1/2-1/2"]

jhlywa:
Thanks @chinanzio .  I'd like to find a way to incorporate translations soon.  I'm going to close this issue.  Keep an eye out for a 1.0.0-rc1 soon that contains the prototype refactor.

PR #111, opened 2016-04-11
Support Loading Arena PGNs

RySchmitt:
Arena includes a space between the move number and the extra two dots to indicate black to move.

For example:
1. e4 e5 (1. .. c5) 2. Nf3


jhlywa:
Please add a test and I'll merge this.

PR #118, opened 2016-06-24
in_checkmate() extended to check which king is checkmated.

sanjay120890:
e.g. chess.in_checkmate('w'); to check white is checkmated,
now we have 3 way to call the same function.
1. chess.in_checkmate(); //true if checkmate
2. chess.in_checkmate('w'); // true if white is checkmated 
3. chess.in_checkmate('b'); // true if black is checkmated

### DIFF
--- a/chess.js
+++ b/chess.js
@@ -732,12 +732,15 @@ var Chess = function(fen) {
     return attacked(swap_color(color), kings[color]);
   }
 
-  function in_check() {
+  function in_check(color) {
+	  if(color){
+		  return king_attacked(color);
+	  }
     return king_attacked(turn);
   }
 
-  function in_checkmate() {
-    return in_check() && generate_moves().length === 0;
+  function in_checkmate(color) {
+    return in_check(color) && generate_moves().length === 0;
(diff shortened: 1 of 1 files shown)


jhlywa:
Thanks for the PR. Instead of adding more to the complexity to the API, I'd rather reuse the existing functions:

```js
if (chess.in_checkmate()) {
    var side = chess.turn() == 'w' ? 'White' : 'Black';
    console.log(side + ' checkmated');
}
```

sanjay120890:
You are right, but it's totally depends on fen and turn, it might be wrong.
See the below link, then you can get the actual problem.
http://www.sanjaynishad.com/MakeAMate/

sanjay120890:
chess.load('rnb1kbnr/pppp1ppp/8/4p3/5PPq/8/PPPPP2P/RNBQKBNR w KQkq - 1 3');
chess.game_over();

This is example we have in read me 
Suppose if user enters b in fen "b KQkq - 1 3" then existing code will return black is checkmated that's wrong.

PR #120, opened 2016-08-16
Attach FEN to history

minnaing:
In the game, when we want to go back to the certain move, we can use history. However, history has not enough information about the situation. History only records **color**, **flags**, **from,** **piece**, **san** and **to**. It is better we have the snippet of **FEN** so that UI can show the situation on that move. So I added FEN to the history. If you think it is useful, here is the pull request. 

_So far, testing fails at History because I changed the history._


jhlywa:
Just cleaning up old issues ... this was implemented as `before` and `after` in v1.0.0. Thanks for the PR

PR #146, opened 2017-03-15
Add an npm script that runs the tests against the minified version.

allonhadaya:


jhlywa:
The minified version was removed from NPM (since most tools minify during bundling these days).

Issue #136, opened 2016-12-31
Promotion callback


barrust:
It would be helpful if one could pass a either a value ('q') or a function to the promotion option in move. This would allow for generating the answer to what to promote to when necessary instead of having to default it to something.

jhlywa:
You must pass the promotion piece when promoting; chess.js doesn't default. You can use one of the following:

```js
var chess = new Chess();

chess.load('8/1P5k/8/8/8/8/8/1K6 w - - 0 1')
chess.move('b8=N')
// -> { color: 'w',  from: 'b7',  to: 'b8',  flags: 'np',  piece: 'p', promotion: 'n',  san: 'b8=N' }

chess.load('8/1P5k/8/8/8/8/8/1K6 w - - 0 1')
chess.move({from: 'b7', to: 'b8', promotion: 'r'});                   
// -> { color: 'w',  from: 'b7',  to: 'b8',  flags: 'np',  piece: 'p', promotion: 'r',  san: 'b8=R' }

chess.load('8/1P5k/8/8/8/8/8/1K6 w - - 0 1');                                                               
chess.move('b7b8=q', {sloppy: true});   
// -> { color: 'w',  from: 'b7',  to: 'b8',  flags: 'np',  piece: 'p', promotion: 'q',  san: 'b8=Q' }          
```

Does this suffice?

barrust:
Sorry, I should have been more clear about the term default. I have been using the second version described while making a UI to help teach my kids about chess. What I was thinking would be a great additon, would be something like this:

``` javascript
var chess = new Chess();
chess.load('8/1P5k/8/8/8/8/8/1K6 w - - 0 1')
chess.move({
    from: 'b7', 
    to: 'b8', 
    promotion: function() {
        var promotion_options = ['q', 'r', 'n', 'b'];
        var promotion_val = promotion_options[Math.ceil(Math.random()*promotion_options.length) - 1];
        return promotion_val;  // this is a random promotion but it could also have been more clever...
    }
}); 
```

Does that make sense? It would allow for the developers of a UI to give the user a choice of promotion pieces after the promotion is signaled.

BillyWM:
Why not just pick a random char from the outside and pass it in? Why does Chess.js need to run a callback for this?

barrust:
The random char function is just to show what I meant by not knowing the value before the move is registered. chess.js does not need to support the callback functionality and I am fine if it is decided that it is not within scope. 

Thanks for your help and a great library

[...]

PR #148, opened 2017-03-22
New Method: is_promotion

allonhadaya:
This pull request is a proposed solution to issue #136 

It solves the underlying problem of enabling clients of chess.js to elegantly handle promotion dialogues.

A proposal is made in that issue to allow `move`'s promotion argument to be a callback. This is an insufficient solution because the callback would have to, well, call back with a promotion choice if determining the promotion choice needs to be done asynchronously. This is the case for graphical prompts that require handling UI events. So now we're talking returning a promise from the callback or invoking a function passed in as an argument. Not to mention that this puts the chess instance in an intermediate state while the callback is making its decision. Very messy.

I believe that `is_promotion` is a more direct solution to the problem since it eliminates the need to handle any asynchronous behaviour in chess.js. Instead the caller can check if the move made is a promotion, prompt the user, and only then `move` the game with the choice. Example:

```js
function handle_user_move(from, to) {

  if (chess.is_promotion(from, to)) {

    // (somehow) prompt the user for their choice, move and promote in the callback
    prompt_promotion_choice(function callback(choice) {
      chess.move({ from: from, to: to, promotion: choice });
    });

  } else {

    // no promotion needed
    chess.move({ from: from, to: to });

  }
}
```


jhlywa:
Hi @allonhadaya. I'm not ignoring your PR's, I've just been really busy lately.  I should have some time to review this in the next few days.

Roofbreaker:
A possible way to solve this problem is check whether a promotion is a valid move and then undo the move before opening any dialog with th user to decide the promotion piece. Something like
        
          // If not, check whether this is a possible promotion requiring further information
          if (boardHandle.chessHandle.move({ from: squareFrom, to: squareTo, promotion: 'q' })!==null){
            boardHandle.chessHandle.undo();
            promotion = true;
            }

allonhadaya:
@jhlywa Oh, thanks! It's no real rush. When you do take a look, I'd be curious to hear what you think of the implementation if there's a better way to do it.

@Roofbreaker You can indicate a promotion choice for non-promotion movement and still get a valid move, so you'd have to then check if the `to` square now had a queen on it. But otherwise, yes, I think that's a valid way to approach the problem: try the move, see if a promotion happened, undo. The current `is_promotion` implementation just mimics a subset of `move` (the move generation step).

Roofbreaker:
@allonhadaya You are right, I did not check what was the result of indicating a promotion in a non-promotion movement. I guess I had no problems so far (I use it in my code) since I check first if the movement without promotion is valid, .i.e. 

        // Check whether the movement is valid
        if (boardHandle.chessHandle.move({ from: squareFrom, to: squareTo})===null){
          // If not, check whether this is a possible promotion requiring further information
          if (boardHandle.chessHandle.move({ from: squareFrom, to: squareTo, promotion: 'q' })!==null){
            boardHandle.chessHandle.undo();
            promotion = true;
            }
          else{
            ....
            }
        }

Nevertheless making sure there is a queen after the promotion movement is probably the safest option.

allonhadaya:
Hey @jhlywa, I was wondering if you might have some free time to take a look at this pull request? Let me know if there's anything that would help you review the code, any changes that should be made, or structural considerations you might want to account for. Thanks!

jhlywa:
Sorry for the long delay @allonhadaya, this fell off my radar.  One suggestion, could you modify your implementation to work with Standard Algebraic Notation as well.  For example `chess.is_promotion('e8=Q')` or even `chess.is_promotion('e7e8Q', { sloppy: true })`.

srensamblador:
What's the status of this PR? Would be very nice to have this functionality in the library

questsin:
Sounds reasonable

kwende:
@jhlywa what's the plan here? What can I do to help get this through?

fabthegreat:
@jhlywa same need here ;)

neofight78:
@jhlywa This PR is super old, but I have seen plenty of PRs / issues wanting to make promotions easier to handle for the UI. This seems like the cleanest API for doing so and has some positive feedback. Do you think it worth adding, if so I  can raise a new PR for it?

jhlywa:
@neofight78 There's an `isPromotion` method on the move object itself, albeit you need a valid move to get access to the method. Let's go ahead and add this (`chess.isPromotion(to, from)`), and add a special note in the README (or new docs!) that it's primarily for UI convenience.

neofight78:
New PR created

PR #149, opened 2017-03-29
Expand docs for .load_pgn

svangordon:
Expanded the documentation for the .load_pgn function. Included additional examples with options.


jhlywa:
Good stuff.  Thanks @svangordon!

PR #151, opened 2017-04-20
Make constants exportable without creating Chess instance

DoumanAsh:
I would like suggest to make constants part of exports aside from Chess instance
It makes a little sense to make constants part of Chess instance.
For backward compatibility i didn't remove it from instance of Chess, but if you're ok with such change i would remove it from instance itself.

I will need to look-up how to properly export things for RequireJS format.
But for global use i think there might be a need for namespace.

Maybe something like: 
```
var CONSTANTS = {
  WHITE: 'b',
  ....
};
```

What do you think?

### DIFF
--- a/chess.js
+++ b/chess.js
@@ -33,21 +33,31 @@
  * https://github.com/jhlywa/chess.js/blob/master/LICENSE
  */
 
+var BLACK = 'b';
+var WHITE = 'w';
+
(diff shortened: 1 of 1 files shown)


jhlywa:
Constants are exportable in newer versions of chess.js.  Thanks for the PR

PR #158, opened 2017-07-08
Less verbose npm script definitions

marcjansen:
The directory `./node_modules/.bin` is always on the PATH when one
runs npm scripts, therefore the references to binaries installed
there can be simplified as follows:

Before:

    ./node_modules/.bin/some-executable

After:

    some-executable

Please review.


jhlywa:
Thanks for the PR!

Issue #161, opened 2017-07-16
Needed disambiguation for pawns even if not necessary


mliebelt:
I have noticed that pawn capture moves are only taken if I include disambigurations. Here is the test code to reproduce it:

    // <reference path='chess.js.d.ts' />
    import { Chess } from 'chess.js'
    describe("Show failing tests for chess.js", () => {
    it ("when needing disambiguration for pawn moves not working", () => {
        var chess = new Chess()
        chess.move("e4")
        chess.move("d5")
        chess.move("xd5")
        console.log(chess.ascii())
    })
    it ("when needing disambiguration for pawn moves working", () => {
        var chess = new Chess()
        chess.move("e4")
        chess.move("d5")
        chess.move("exd5")
        console.log(chess.ascii())
    })
    })

The result when running that test case is the following:

        Started
       +------------------------+
     8 | r  n  b  q  k  b  n  r |
     7 | p  p  p  .  p  p  p  p |
     6 | .  .  .  .  .  .  .  . |
     5 | .  .  .  p  .  .  .  . |
     4 | .  .  .  .  P  .  .  . |
     3 | .  .  .  .  .  .  .  . |
     2 | P  P  P  P  .  P  P  P |
     1 | R  N  B  Q  K  B  N  R |
       +------------------------+
         a  b  c  d  e  f  g  h
    
    .   +------------------------+
     8 | r  n  b  q  k  b  n  r |
     7 | p  p  p  .  p  p  p  p |
     6 | .  .  .  .  .  .  .  . |
     5 | .  .  .  P  .  .  .  . |
     4 | .  .  .  .  .  .  .  . |
     3 | .  .  .  .  .  .  .  . |
     2 | P  P  P  P  .  P  P  P |
     1 | R  N  B  Q  K  B  N  R |
       +------------------------+
         a  b  c  d  e  f  g  h
    
    .
    
    
    2 specs, 0 failures
    Finished in 0.029 seconds

As you can see, only in the second case the pawn move is done. It did not work with the valid notation "xd5".

mliebelt:
Reading the specification again (http://www.saremba.de/chessgml/standards/pgn/pgn-complete.htm#c8.2.3), I have found my fault:

> 8.2.3.3 Basic SAN construction: ... Capture moves are denoted by the lower case letter "x" immediately prior to the destination square; pawn captures include the file letter of the originating square of the capturing pawn immediately prior to the "x" character.

[...]

PR #162, opened 2017-08-03
#163 Reimplemented sloppy mode (move_from_san) with more flexibility

mliebelt:
Tried to come up with a solution. All unit tests are running, I have added 2 unit tests (which use the sloppy mode), that are then working as well.

### DIFF
--- a/chess.js
+++ b/chess.js
@@ -1070,47 +1070,70 @@ var Chess = function(fen) {
   function move_from_san(move, sloppy) {
     // strip off any move decorations: e.g Nf3+?!
     var clean_move = stripped_san(move);
-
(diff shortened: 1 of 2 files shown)


mliebelt:
After having tried first to implement #161, I saw that this is only a special case from the more general, that allows all kinds of variations when reading a PGN move. So I rewrote the sloppy code, and fixed all unit tests after that.

axolotl-logic:
Thank you for you contributions! Unfortunately, time has been unfairly cruel to this PR and it has grown dusty - I apologize for that. We are going through and closing out old PRs and this is on that list. If you are still interested in contributing to the code base, please comment and accept my apologies for closing this PR. Attention: @neofight78

PR #166, opened 2017-11-11
Added fens() to expose internal fens array which is the point of this fork.

sandy98:
The whole idea is that the game object keeps internal state not only of movements (current behavior) but of every position reached, including the original. The point is that board widgets deal with positions, rather than moves, so having positions (fens) exposed, allows code like this (example from my https://github/sandy98/next-chess-board): function load_pgn(pgnStr) {
  var result = this.game.load_pgn(pgnStr);
if (result) {
   this.setState({currentPosition: 0, positions: this.game.fens()});
}
Pretty straightforward.
Memory footprint is absolutely low, and performance keeps blazingly fast.


jhlywa:
Closing this in favor of #120

PR #167, opened 2017-11-21
Add typescript type definitions

captainsano:
Add Typescript type definitions alongside the package. Also, a test script is included to exercise the types.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,124 @@
+declare module 'chess.js' {
+  type FEN = string
+  
+  type Color = 'w' | 'b'
+  
+  type PieceIdentifier = 'k' | 'q' | 'r' | 'n' | 'b' | 'p'
+  type PromotionPieceIdentifier = 'q' | 'r' | 'n' | 'b'
+  
+  interface Piece {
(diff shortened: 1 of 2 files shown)


Nicals:
Some type definition are already done here: https://www.npmjs.com/package/@types/chess.js

zacnomore:
[From the typescript documentation:](http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html)
> Publishing
> 
> Now that you have authored a declaration file following the steps of this guide, it is time to publish it to npm. There are two main ways you can publish your declaration files to npm:
> 
>     bundling with your npm package, or
>     publishing to the @types organization on npm.
> 
> If your package is written in TypeScript then the first approach is favored. Use the --declaration flag to generate declaration files. This way, your declarations and JavaScript will always be in sync.
> 
> If your package is not written in TypeScript then the second is the preferred approach.
> Including declarations in your npm package

Noting the last line, unless this is being rewritten in TypeScript I think it's already in the correct location

neofight78:
chess.js now uses TypeScript

PR #194, opened 2018-10-16
Added attribute toDo in move() function

ZaninDavide:


questsin:
What’s is the purpose of toDo?

ZaninDavide:
The idea was to create a way to predict the consequences of a move. The fastest way I found is adding this parameter(toDo) but the best one would probably be adding a completely new function.
I'm writing a chess application and I want to know what would be the san of the move(a to b) if the player did this move. The way I did it is duplicating the game and doing the move(also possible doing the move and then come back) but it's not very efficient.

I hope I was clear.

Issue #198, opened 2018-12-19
Build failing for older Nodejs Versions


pscott-au:
![image](https://user-images.githubusercontent.com/5900187/50194363-34eda000-0386-11e9-8c38-afc275ea6d44.png)

jhlywa:
Fixed in 8e854c7

PR #199, opened 2018-12-19
Remove older failing NodeJS Versions

pscott-au:
https://github.com/jhlywa/chess.js/issues/198


jhlywa:
Fixed in 8e854c7004c391f150e10fad04568d34f248bc9e

PR #235, opened 2020-05-03
add support for annotations

jtwires:
* add methods for creating/retrieving/removing annotations
* parse bracket- and semicolon-style annotations when loading PGNs
* output bracket-style annotations when producing PGNs

### DIFF
--- a/README.md
+++ b/README.md
@@ -635,6 +635,99 @@ chess.validate_fen('4r3/8/X12XPk/1p6/pP2p1R1/P1B5/2P2K2/3r4 w - - 1 45')
 //     error: '1st field (piece positions) is invalid [invalid piece].' }
 ```
 
+### .get_comment()
+
+Retrieve the comment for the current position, if it exists.
+
+```js
+const chess = new Chess()
+
+chess.load_pgn("1. e4 e5 2. Nf3 Nc6 3. Bc4 Bc5 {giuoco piano} *")
+
+chess.get_comment()
+// -> "giuoco piano"
+```
+
+### .set_comment(comment)
+
+Comment on the current position.
+
+```js
+const chess = new Chess()
+
+chess.move("e4")
+chess.set_comment("king's pawn opening")
+
+chess.pgn()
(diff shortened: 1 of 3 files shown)


jhlywa:
Thanks Jake.  I'll review this over the next few days.

jhlywa:
Thanks @jtwires.  It's clear you put a lot of work into this PR.  I've got a few questions and suggestions before I merge.

1.  I can see the use-case arising where a user would want to remove all annotations from a PGN.  How would you feel about adding an explicit `clear_annotation()` function that could be called with an optional argument to clear all annotations (e.g. `clear_annotation({target: 'game'})` or `clear_annotation({all: true})` or something similar, picking good names is hard).  Returning the annotation that was cleared (or null) would be a nice touch too.

2. The [PGN spec](https://www.chessclub.com/help/PGN-spec) refers to these annotations as comments; basically anything in `{ ... }` or `;...`, while using the term annotation for Recursive Annotation Variations.  Could you change the API to use the term `comments` instead?

3.  I've added a few inlines comments in the PR itself.

Many people have been requesting this capability.  Thanks again for your contribution.

README.md
```diff
@@ -58,6 +58,61 @@ const chess = new Chess(
...
+chess.pgn()
+// -> "1. e4"
+```
+
+### .annotation()
+
+Retrieve the annotation for the current position, if it exists.
+
+```js
+const chess new Chess()
```
jhlywa, line 86: Minor type `const chess = new Chess()`

chess.js
```diff
@@ -1541,14 +1615,46 @@ var Chess = function(fen) {
         }
       }
 
+      var to_hex = function(string) {
```
jhlywa, line 1618: Can you add a comment of too here to explain what you're doing with hex and URI encoding?  It's not immediately clear to me.  Thanks

jtwires:
cool, `comment` it is..

jhlywa:
0.11.0 has been pushed to npm.  I'll gen up a changelog sometime over the weekend.

PR #239, opened 2020-06-21
Specific id for each piece

ZaninDavide:
I made so that each piece is created with a unique id that keeps track of the piece through the game. I wrote new tests for this feature.
IDs can be very useful for applications based on ReactJS: allow to give correct keys to the pieces (improving performance) and to easily create animations.

Practical changes:
```js
chess.get('a2') // -> { type: 'p', color: 'b', id: 16 }
```
```js
chess.history({ verbose: true }) 
/*
 -> [{ color: 'w', from: 'e2', to: 'e4', flags: 'b', piece: 'p', san: 'e4' },
     { color: 'b', from: 'e7', to: 'e5', flags: 'b', piece: 'p', san: 'e5' },
     { color: 'w', from: 'f2', to: 'f4', flags: 'b', piece: 'p', san: 'f4' },
     { color: 'b', from: 'e5', to: 'f4', flags: 'c', piece: 'p', captured: 'p', san: 'exf4', captured_id: 21 }]
*/
```
```js
chess.indices()
/*
   +----------------------------------------+
   8 | 000  001  002  003  004  005  006  007 |
   7 | 008  009  010  011  012  013  014  015 |
   6 | ...  ...  ...  ...  ...  ...  ...  ... |
   5 | ...  ...  ...  ...  ...  ...  ...  ... |
   4 | ...  ...  ...  ...  ...  ...  ...  ... |
   3 | ...  ...  ...  ...  ...  ...  ...  ... |
   2 | 016  017  018  019  020  021  022  023 |
   1 | 024  025  026  027  028  029  030  031 |
     +----------------------------------------+
        a    b    c    d    e    f    g    h
*/
```

Retro-compatibility is preserved and all tests run fine.


neofight78:
Thank you for the PR. Obviously this is quite old now, but also I'm not sure adding an id to the core logic can be justified for such a specific use case. Whilst perhaps convenient, I'm not sure it's even required for implementing an efficient react front end. For example using the square as an id is generally sufficient, although I am sure there must be other viable options too.

PR #245, opened 2020-08-24
using invariate fen for repetition counting.  Using clones of board a…

jvsteiner:
…nd history to remove reliance on undoing and redoing global state.  Added tree based structure to represent history instead - this will eventually allow us to support variations.

Hi - this is not an actual pull request, in the sense that I think it would be a good idea to blindly accept it - rather consider it a sort of sketch showing where I think this library should move towards.  I have been using you library for the last few weeks and noticed a couple of things that I found important and wanted to demonstrate to you.

1. I discovered that there is a deep reliance on global shared state in `make_move` and `undo_move`  for all manner of calculations, fen generation, move feasibility etc that have nothing to do with recording an actual move in the actual game being played/recorded/analyzed.
2. You allow to `undo_move` but not to `redo_move` - this is again related to global shared state, because to undo a move, you have to pop the last move, and get rid of it.  there is therefore no distinction between moving backwards in a line/game/variation and literally erasing it from memory.
3. No variations are possible in the current structure because of (2) above
4. example - threefold repetition checking involves literally playing back through a bunch of moves to see if there was a repetition, which is grossly inefficient.
5. There's other stuff, like the fen validation, which is noted in the source code already

I'd be willing to contribute a bit of time to try and improve things, but first I wanted to check with you and see if you are open to this, and available to help, because frankly there would be a lot of stuff to refactor, and it would be easier if you, as the original author, were involved enough to answer questions, etc. so I would not have to reverse engineer things unnecessarily. 

I'd like to make it more like the excellent `python-chess` in both architecture and API.

What do you think?

-Jamie

P.S. My motivations are that I am considering making a chess website, and I'd like to have a nice library, with a liberal license that I can use.  I could just do this for my own purposes, but I see that there isn't another really great library available, and as a chess fan, and developer, that feels wrong in 2020.

### DIFF
--- a/chess.js
+++ b/chess.js
@@ -150,14 +150,133 @@ var Chess = function(fen) {
     ]
   }
 
-  var board = new Array(128)
-  var kings = { w: EMPTY, b: EMPTY }
-  var turn = WHITE
-  var castling = { w: 0, b: 0 }
-  var ep_square = EMPTY
-  var half_moves = 0
-  var move_number = 1
-  var history = []
+  // var Test = function() {
+  //   return {
+  //     a:out,
+  //     test: function() {
+  //       console.log(this)
+  //     }
+  //   }
+  // }
+
+  var GameNode = function(move, parent) {
+
+    if (typeof move === 'undefined') {
+      move = null;
+    }
+
+    if (typeof parent === 'undefined') {
+      parent = null;
+    }
+
+    return {
+      move: move, // should be "ugly" move leading to this node, or null if this is the root
+      kings: { b: kings.b, w: kings.w },
+      turn: turn,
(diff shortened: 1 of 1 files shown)


atinm (commented):
Maybe what you need is a stack of "contexts" (GameNodes?) and always pass a context to all functions so there really is no "global board" etc. Then when adding a variation, you push the the current_context on the stack and start working on a clone of the current context as your new current_context. When you end the variation, you pop the top of the stack context into your current_context. Same thing for doing the other move/undo_move dances in other places in the code.

neofight78:
I think something similar to what you are talking will be needed for RAV support, see #522

PR #253, opened 2020-12-16
Added Chess 960 support

atinm:
Chess 960 support, with tests.


atinm:
I just saw that there is another PR also trying to solve this and that is also pending merge or even comments, so  maybe this repo is not really getting much attention.

PR #254, opened 2020-12-17
Has checkmate

topce:
add function to check if there is checkmate (aka checkmate in one move)

### DIFF
--- a/chess.js
+++ b/chess.js
@@ -1406,7 +1406,7 @@ var Chess = function (fen) {
       )
     },
 
-    has_checkmate: function() {
+    has_checkmate: function () {
       return has_checkmate()
     },
 
--- a/chess.js
+++ b/chess.js
@@ -879,6 +879,22 @@ var Chess = function(fen) {
     return repetition
   }
 
+  function has_checkmate() {
+    var moves = generate_moves()
+    var winning_move = null
+    for (let index = 0; index < moves.length; index++) {
+      const move = moves[index]
+      make_move(move)
+      if (in_checkmate()) {
+        winning_move = move
+        undo_move()
+        break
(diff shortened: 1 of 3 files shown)


chess.js
```diff
@@ -879,6 +879,22 @@ var Chess = function(fen) {
...
 
+  function has_checkmate() {
+    var moves = generate_moves()
+    var winning_move = null
+    for (let index = 0; index < moves.length; index++) {
+      const move = moves[index]
+      make_move(move)
+      if (in_checkmate()) {
+        winning_move = move
+        undo_move()
```
atinm, line 890: Remove this, you already have an undo_move call outside the if statement so you'll do an extra undo if you find a checkmate.

topce: Hi 
> Remove this, you already have an undo_move call outside the if statement so you'll do an extra undo if you find a checkmate.
thank you for code review.
You are right from performance point of view 
however I wanted  method not to modify chess object 
after calling method same side need to be on turn.
 
Also I return just one winning move : first found
potentially I could return array of  moves , if there are more 
moves to checkmate in one .  
potential improvement.
What do you think?

atinm: Yes, I just saw that you have a break that breaks out of the for loop - code is fine! I don't know when/if this will get merged though - I am not sure this repo is maintained!

topce: me neither, but this is my 2 cents  ;-)

ianasheu:
moves() already give a move with a #

topce:
@ianasheu ,@utzig26 yes good point  but it does not give you checkmate in 2 , 3 , 4
brute force checkmate 
this PR could be generalized to do this.

PR #258, opened 2021-01-13
Fix chess.js import

Sigmus:


neofight78:
Fixed in newer versions which use import rather than require

PR #264, opened 2021-02-20
making blank line after header optional

nextchessmove:
Greetings, and thank you for the excellent library!  We use it at nextchessmove.com.  A few users have complained about not being able to paste PGN into the app and website, and in all cases thus far it has been because of a lack of a blank line separating the header and movetext sections.

Would you be open to open to updating chess.js to parse malformed PGN of this type?  This is my initial stab at it.  Note that with this change, it parses the malformed PGN regardless of the sloppy setting -- I'm not sure if that's right.

### DIFF
--- a/__tests__/chess.test.js
+++ b/__tests__/chess.test.js
@@ -741,6 +741,16 @@ describe("Load PGN", function() {
       expect: true,
       sloppy: true
     },
+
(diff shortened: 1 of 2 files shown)


neofight78:
No longer an issue in recent versions

PR #271, opened 2021-03-24
Speed up gen moves - ~30% speed up - Test Cases from 6.8s to 4.5s

ejezisek:
The purpose of this is to speed up the `generate_moves` function within `move_from_san`.  Specifically; if a piece type is supplied then we can significantly speed this up by only checking for valid moves of that piece.  When dealing with a lot of `pgns` like in an opening database of sorts this gives a decently significant speed up.  There is more that can likely be done; but regardless the test cases ran on my machine on master completed in 6.8 seconds and the test cases on this completed in 4.5.

### DIFF
--- a/chess.js
+++ b/chess.js
@@ -544,7 +544,10 @@ var Chess = function(fen) {
       typeof options !== 'undefined' && 'legal' in options
         ? options.legal
         : true
-
+      var piece_type = typeof options !== 'undefined' && 'piece' in options
+        ? options.piece.toLowerCase()
+        : true
+    
     /* are we generating moves for a single square? */
     if (typeof options !== 'undefined' && 'square' in options) {
       if (options.square in SQUARES) {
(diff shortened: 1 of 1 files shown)


ajax333221:
At a quick glance I see that it looks for the first char if its an `a-h` then if not assume the first char will give the piece. But are you sure this would work for castling to detect the king.

ejezisek:
> At a quick glance I see that it looks for the first char if its an `a-h` then if not assume the first char will give the piece. But are you sure this would work for castling to detect the king.

> node
> chess.load_pgn('1. e4 e5 2. Nf3 f6 3. Bc4 Ne7 4. O-O')
true
> chess.fen
[Function: fen]
> chess.fen()
'rnbqkb1r/ppppn1pp/5p2/4p3/2B1P3/5N2/PPPP1PPP/RNBQ1RK1 b kq - 3 4'

PR #300, opened 2021-11-14
Allow deleting headers

JHK:
This makes it possible to unset a header.


JHK:
Obviously it is also possible to add a method `clear_header(header: string)` or `remove_header(header: string)`. I think explicitly setting a `undefined` value would meet peoples expectations already when they try to unset a header. Thus I went for this simple approach.

neofight78:
Recent versions have .removeHeader(field: string): boolean

Issue #259, opened 2021-01-30
generate uci moves for use with engines like stockfish


honkskillet:
The common uci protocol users a type of long algebraic notation for moves.

> //Sending to Stockfish
> position startpos moves e2e3 d7d6 d1f3 b8d7 f1c4 e7e6 d2d4 d6d5
> 

It would be nice to be able to have a uciMove field on move objects.  As well as being able to get the full list of moves in UCI format.

I think this would generate a uci move from the current verbose move object.

> export function getLongAlgebraicNotationFromMoveObject (moveObj) {  
>   return (moveObj.promotion) 
>       ? moveObj.from+moveObj.to+moveObj.promotion //when there is a promotion
>       : moveObj.from+moveObj.to;
> }

jhlywa:
Added a `lan` field (Long Algebraic Notation) to move objects in master.

PR #303, opened 2021-12-01
added move_to_gan() function which converts a move into Grid Algebraic Notation

RuthShryock:
useful for generating moves with chess engines

### DIFF
--- a/chess.js
+++ b/chess.js
@@ -742,17 +742,11 @@ var Chess = function (fen) {
      */
  function move_to_gan(move) {
   var output = ''
-  if (move.flags & BITS.KSIDE_CASTLE) {
-    output = 'O-O'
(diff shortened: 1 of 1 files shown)


RuthShryock:
resolves #259

ajax333221:
> resolves #259

UCI move castling is like for example: **e1g1**, but the `move_to_gan()` seems to use: **O-O**

ajax333221:
now that it is basically move_to_uci() shouldn't it be named like that? (just a thought, I don't know hehe)

(-:

RuthShryock:
yes thanks :)

eiannimaersk:
You should add the new method into the documentation as well so it's easy to find for new users.

jhlywa:
This was recently added to master under the `lan` (Long Algebraic Notation) field.

PR #310, opened 2022-01-13
Improve moves list results

ianasheu:
First, the evolution simply provides the moves of the opponent's pieces. For this, a new option is added to the .moves() function: {opponent: true}. Then, the function also returns defensive moves with the option: {defensive: true}. By combining the 2 options, the function returns the defensive moves of the opponent. The option: {verbose: true} is still functional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -571,10 +571,16 @@ chess.moves({ verbose: true })
 //     },
 //     ...
 //     ]
+
+chess.moves({ opponent: true })
+
+chess.moves({ defensive: true })
 ```
 
 The _piece_, _captured_, and _promotion_ fields contain the lowercase
 representation of the applicable piece.
+
+In case of defensive option, _captured_ indicates the type of the defended piece
 
 The _flags_ field in verbose mode may contain one or more of the following values:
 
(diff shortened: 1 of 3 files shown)


ianasheu:
try here: http://ianasheu.free.fr/chess/

ianasheu:
troubles with disambiguated san in opponent moves to fix

ianasheu:
troubles with disambiguated san in opponent moves to fix

ianasheu:
to improve : the king is not counted in the defenders if the piece is threatened, even if there is only one threat

ianasheu:
to improve : the king is not counted in the defenders if the piece is threatened, even if there is only one threat

ianasheu:
to improve : the king is not counted in the defenders if the piece is threatened, even if there is only one threat

ianasheu:
to improve : the king is not counted in the defenders if the piece is threatened, even if there is only one threat

ianasheu:
-write tests for defensive moves
-enumerate limitations

ianasheu:
-update tests file from new master version...
-write tests for defensive moves
-enumerate limitations

PR #312, opened 2022-02-20
Include square to board() output

fitztrev:
This PR adds the square notation to the `board()` output.

I was looking for an easy way to find all the squares that a certain type of piece would be on.


jhlywa:
@fitztrev  This change looks reasonable and should be backwards compatible with any existing code that uses the `.board()` function. The tests, however, appear to be failing in CI though. Can you tweak the tests and docs before I merge?  Thanks

jhlywa:
Thanks Trevor!

PR #318, opened 2022-03-22
Convert package to standard JS modules

### LINKED ISSUE #246: Publish standard modules to npm
It would make it a lot easier to consume this library from standard JS modules if it were also published as JS modules.

Module support is extremely good across current JS environments. All major browsers have supported modules for years, and Node has supported them in the last 3 major versions, and soon to be the last 4 majors and last 2 LTS's. All bundlers also automatically support consuming dependencies published as modules and will convert them to whatever output format the user chooses.

justinfagnani:
Fixes #246 

**Note**: This is definitely a major breaking change!

There are ways to make this less of a breaking change and publish both CommonJS and JS modules, but this PR contains the simplest way to publish modules if you can ask your users to upgrade. It is of course important to acknowledge that importing standard JS modules into CommonJS is difficult and requires using dynamic `import()` or a bundler.

I split this into two main commits. The first simply exports the `Chess` function as-is. The second reorganizes some things to be a little more idiomatic for modules and puts constants in the top-level scope of the module. This means that `KING`, `PAWN`, etc can be imported like:

```ts
import {Chess, KING} from 'chess.js';
```

which solves a question previously in the code about how to best export public constants.

I also considered converting `Chess` to a real class, but am saving that for later if you're interested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,8 @@ yarn add chess.js
 The code below plays a random game of chess:
 
 ```js
-const { Chess } = require('./chess.js')
+import { Chess } = from 'chess.js';
+
 const chess = new Chess()
 
 while (!chess.game_over()) {
--- a/__tests__/chess.test.js
+++ b/__tests__/chess.test.js
@@ -1,6 +1,4 @@
-if (typeof require != 'undefined') {
-  var Chess = require('../chess').Chess
-}
+import { Chess } from '../chess.js'
 
 describe('Perft', () => {
   const perfts = [
--- a/__tests__/chess.test.js
+++ b/__tests__/chess.test.js
@@ -1,4 +1,4 @@
-import { Chess } from '../chess.js'
+import { Chess, WHITE, BLACK, PAWN, KNIGHT, BISHOP, ROOK, KING, QUEEN, SQUARES } from '../chess.js'
 
 describe('Perft', () => {
   const perfts = [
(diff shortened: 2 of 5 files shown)


jhlywa:
Thanks @justinfagnani!  Would it be helpful to publish this as 1.0.0-alpha?  I'd like to make a few API changes prior to 1.0.0 and I'm unsure if changing the API between 1.0.0-alpha and 1.0.0 is bad form.

jhlywa:
I'll need to spend some time investigating the impacts of this change wrt the community. chess.js has been major version 0.x.x for 13 years (which is an eternity in JS years - the coding style is a dead giveaway), so I think the users are expecting a stable API even though it's major version 0.  Adhering to the principle of least surprise, does npm/yarn prevent minor version updates for caret deps if they're pre-1.0.0?  As long as the upgrade to 0.13.0 is explicit, I'm definitely in favor of merging this and cutting a new release.

I've been working on a private typescript rewrite on and off (honestly, mostly off) for the past year. The current typescript branch was a first attempt, but it's more of a false start, so please ignore. I should really push my latest, publish a roadmap, and seek some feedback/input/PR's before going too much further.

jhlywa:
I haven't verified this, but pre-1.0.0 caret updates appear to be explicit (https://docs.npmjs.com/cli/v6/commands/npm-update#caret-dependencies-below-100). I'll try to confirm it with verdaccio this weekend.

Additionally, a basic roadmap for 1.0.0 has been published at #319

jhlywa:
The version upgrade pre-1.0.0 is explicit (when crossing minor versions), so this upgrade should be safe. v0.13.0 has been published to npm. Thanks for the PR!

GabrielDelepine:
`type: "module"` has been removed from `package.json` in this commit https://github.com/jhlywa/chess.js/commit/37ac2aff7b5e945ee0d657016b1d5126d653ad01#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

Please re-consider @jhlywa 

(Ideally, you will want to adopt the `exports` field)

jhlywa:
Ah, good catch @GabrielDelepine . Thank you

I'm unfamiliar with the `exports` field.  I'll need to read up a bit first before adding

GabrielDelepine:
Thanks for the quick fix @jhlywa 
But be aware that the current JS published in NPM as `commonJs` instead of `module`!

I opened a PR to fix it https://github.com/jhlywa/chess.js/pull/374

GabrielDelepine:
I totally understand your decision. ESM is giving birth in pain due to some incompatibilities with CommonJS.

The `exports` field of `package.json` is the best way to go to have a dual CommonJS + ESM package but I totally understand if it's not on top of your todo list right now

Issue #319, opened 2022-04-01
Roadmap: chess.js 1.0.0


jhlywa:
After 13 years of pre-1.0.0 development .... it's time.

This is a working document to track planned features/changes and their statuses for version 1.0.0.  Please comment if there's something you'd like to see included or changed. I appreciate and value your feedback.  Check back frequently for updates.

I'll push a basic typescript implementation to **dev** branch to use as starting point. 

~~Please wait for the **dev** branch before submitting PR's.~~

**dev** branch is now active

## Proposed 1.0.0 Changes

- [ ] Code
  - [x] Rewrite library in typescript
  - [x] Use camelcase for API and everything else
    - [x] enforce with eslint rule
  - [x] Use exceptions in error states instead of returning `null`. This will allow the library to provide more detailed error messages for common issues (e.g. bad moves, bad FEN, bad PGN).  The functions that throw exceptions are listed in the API Changes section.
    - [x] throw exception for bad FEN in `.load()` 
    - [x] throw exception for bad FEN  in `Chess()` constructor
    - [x] throw exception for bad PGN in `loadPgn()`
    - [x] throw exception for bad SAN in `move()`
    - [x] throw exception for bad move object in `move()`
  - [x] Switch to using the sloppy FEN/PGN parser by default.  The phrase sloppy is kind of passive aggressive, so maybe use the terms `permissive` and `strict`.
    - [X] `load` in permissive mode (the default mode) should let the user load FEN without supplying castling rights, ep square, and move numbers (e.g. chess.load(`rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w`))
    - [x] `load` and `loadPgn` should allow an option to use the strict parser 
  - [x] `moves()` should work even if there are no kings on the board. 
  - [x] Only set the en passant square when the opponent can legally make an en passant move. (see #294)
  - [x] Add a 'lan' (better name or is this ok??) property for long algebraic notation (see #259)
  - [x] Improve FEN validation.  It should detect common hand-entered such as missing kings and incorrect castling rights. 
  - [ ] Use fast-check for property-based testing
  
- [ ] Docs
  - [ ] create a chess.js logo (maybe this is something the community can vote on??)
  - [ ] Move documentation to chessjs.org 
  - [ ] Add an FAQ section to answer common questions

- [ ] Other
  - [ ] Github issue template ensuring the reporter know the rules of chess before filing a bug  (en passant gets flagged as a bug a few times a year) 
[...]

PR #320, opened 2022-04-04
Add in_double_check()

fitztrev:
I was looking for a way to be able to detect "double checkmates" ([example here](https://lichess.org/E92ldwZr/black#30))

I'll submit this as a draft PR for discussion, knowing that #319 will have a new development branch. If this gets approved, I will re-do against the `FIX_ME` branch.

### DIFF
--- a/__tests__/chess.test.js
+++ b/__tests__/chess.test.js
@@ -136,6 +136,30 @@ describe('Single Square Move Generation', () => {
   })
 })
 
+describe('Double Check', () => {
+  test.each([
+    '1n6/8/8/3k4/5N2/8/8/K2R4 b - - 0 1',
+    '1n6/4k3/8/5N2/1B6/8/8/K7 b - - 0 1',
+    '8/4k3/5q2/8/8/3n4/5K2/8 w - - 0 1',
+  ])('Testing double check: %p', (fen) => {
+    const chess = new Chess(fen)
+    expect(chess.in_check()).toBe(true)
+    expect(chess.in_double_check()).toBe(true)
+  })
+})
+
+describe('Double Checkmate', () => {
+  test.each(['k4qqq/8/8/8/8/4n3/6K1/8 w - - 0 1'])(
+    'Testing double check: %p',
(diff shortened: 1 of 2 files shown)


jhlywa:
Hey Trevor, could you do the same thing with something a little more generic?  Maybe simplify the concept into an `attackers()` API call?  In the end it's a few more lines of code for your specific use case, but I think it results in greater flexibility for the library.  Something like:
```
const kingSquare = somehow get the king square for the side to move 
const attackers = chess.attackers(kingSquare) /// [{piece: 'n', square: 'f3', color: 'b'}, ...]
const doubleAttack = attackers.length === 2
```

So we'd need a way to retrieve the square for the king, or more generically any piece of either color. Maybe return a list of squares for the piece/color in question.

This approach also gives us `defenders()` for free too (something like `function defenders(square) { return attackers(square, swap(this.turn)) }`), right?

Thoughts?

fitztrev:
I like that approach.

In this example, would the bishop, queen, and e1 rook count as attackers of e5? And if so, would that work with the way square/attack mappings are currently setup? I'll have to dig into that part of the code more to understand how it works.

image

jhlywa:
Maybe something like `attackers('e4', { xray: true })` for something like that?

Xray gets a little complicated if some of the attackers are pinned. I guess the pinned piece counts as an attacker but nothing "behind" it does? I'm ok punting on xray support for the time being.

fitztrev:
I pushed some updates. Below would be a proposed sample use:

Given this position:
Screen Shot 2022-04-08 at 11 57 32 PM

```js
// get white attackers of d5
chess.attackers('w', 'd5')

// result
[
  { square: 'f3', color: 'w', type: 'q' },
  { square: 'd1', color: 'w', type: 'r' },
]
```

JacobEvelyn:
A project I'm working on would really benefit from this `attackers()`/`defenders()` API. (The `xray` feature would be nice but is not mandatory for me.) Let me know if there's anything I can do to help get this polished and merged.

PR #327, opened 2022-04-30
Fix a bug in validate_fen.

eleurent:
When the move number field is empty (trailing space at the end), the validation returned valid since `isNaN('')` is false and `(parseInt('') <= 0)` is `NaN <= 0`, which is false.

Added the new (previously failing) test case.


jhlywa:
Thanks for the PR @eleurent.

PR #332, opened 2022-06-02
loagPgn exceptions

swaltek:
`loadPgn` now throws and exception instead of just returning false

Change mention in #319
I'd be happy to make any necessary changes

I also think it would be helpful if `loadPgn` gave a warning if a SetUp tag was present without a FEN tag and vice versa. If this gets pulled I'd love to add that as well.


jhlywa:
Thanks @swaltek.  I'll try to get this reviewed this weekend.

PR #348, opened 2022-09-27
New material count feature.

Matthew-Nicholson:
I recently used this feature in an application to select interesting positions to analyse and filtering out positions where material was not close to equal without requiring the use of a chess engine. 

Perhaps it would be a useful addition to chess.js? 

This is my first PR so please look it over carefully and I'll be happy to fix anything if you decide this is a feature you want.

PR Includes "material" function, tests, and an updated readme. 

Thanks,
Matt
matthewn1991@gmail.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -623,6 +623,10 @@ chess.pgn({ max_width: 5, newline_char: '<br />' })
 // -> '[White "Plunky"]<br />[Black "Plinkie"]<br /><br />1. e4 e5<br />2. Nc3 Nc6'
 ```
 
+### .material()
+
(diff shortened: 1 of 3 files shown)


LouieMartin:
I think doing it something like this is better (what im currently using):
```
const evaluate = () => {
  const result = {
    black: 0,
    white: 0,
  };

  board.forEach(piece => {
    if (piece && piece.type in pieceValues) {
      if (piece.color === 'w') {
        result.white += pieceValues[piece.type];
      } else {
        result.black += pieceValues[piece.type];
      }
    }
  });

  return result;
};
```

LouieMartin:
PS - chess.board is not an actual value I just added `, board` on line 1736

PR #350, opened 2022-10-20
Add postinstall build call

KrzysiekWyka:
Adding important `postinstall` hook to build js files.


jhlywa:
Thanks @KrzysiekWyka!

PR #354, opened 2022-11-26
fix typo inCheckmate to isCheckmate

kevo1ution:


kevo1ution:
fixes #357

jhlywa:
Thanks for the PR.  The plan for 1.0.0 is to have an "in" alias for all isXXX functions (e.g. inStalemate would be an alias for isStalemate).

PR #358, opened 2022-12-16
Better types for verbose options

Angshu31:
TypeScript complains every time I try to use `moves()` and assume that the result only contains strings.
This PR fixes that issue by using a type parameter to determine the return type

### DIFF
--- a/src/chess.ts
+++ b/src/chess.ts
@@ -905,16 +905,18 @@ export class Chess {
     return this.isCheckmate() || this.isStalemate() || this.isDraw()
   }
 
-  moves<T extends boolean = false>({
-    verbose = false,
+  moves<T extends boolean | undefined = undefined>({
+    verbose,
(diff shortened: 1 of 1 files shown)


robotkutya:
:+1: this is very annoying.

But instead of returning `any`-s and passing in `type` we should simply narrow the type, no?

So when `{verbose: false}` we return with `string` and when `{verbose: true}` we return with `Move`.

Angshu31:
I'm guessing you mean overloading the function? That's actually a good idea, I'm surprised I didn't think of that. I'll use that instead 👍

jhlywa:
@Angshu31 Thanks for the PR.  Happy to merge these in, but first can you please incorporate @robotkutya's suggestions?

jhlywa:
Thanks all for the feedback. This issue has been fixed in master (w/ overloads).

PR #366, opened 2022-12-26
fix loadPgn() to handle pgn without any moves

### LINKED ISSUE #362: loadPgn doesn't work on a pgn without any moves yet
**What**
`chess.loadPgn` doesn't work on a pgn that has no moves yet.

```js
const { Chess } = require('chess.js')

const game = new Chess();
game.header('White', 'id0');
game.header('Black', 'id1');
// game.move('e4'); // uncomment this for loadPgn to work

console.log(game.pgn())
/** output:
[White "id0"]
[Black "id1"]

*/

console.log('trying to load into another game instance')

const gameCpy = new Chess();
const result = gameCpy.loadPgn(game.pgn())
console.log('result:', result) // output: false
```

**Why**
It would be nice to rely solely on the pgn to describe the game, especially when saving the string in a database to track the game to be loaded for future play/moves.

kevo1ution:
fixes #362 

**What**
- handles empty move text by matching headers when move text is empty and by parsing out moves that are empty strings

**Why**
- loadPgn() should handle a PGN that does not contain any moves and only headers

### DIFF
--- a/__tests__/load-pgn.test.ts
+++ b/__tests__/load-pgn.test.ts
@@ -36,7 +36,7 @@ describe('Load PGN', () => {
         '[Round "6"]',
         '[White "Kevo1ution"]',
         '[Black "Billy Bob Jr."]',
-        '',
+        '', // extra newlineChar is expected because pgn() function adds it
       ],
       expect: true
     },
--- a/__tests__/load-pgn.test.ts
+++ b/__tests__/load-pgn.test.ts
@@ -41,7 +41,7 @@ describe('Load PGN', () => {
       expect: true
     },
     {
-      pgn: [''],
+      pgn: [''], // extra newlineChar is expected because pgn() function adds it
       expect: true
     },
     {
--- a/__tests__/load-pgn.test.ts
+++ b/__tests__/load-pgn.test.ts
@@ -29,6 +29,17 @@ describe('Load PGN', () => {
       ],
       expect: true,
     },
+    // GitHub Issue #362 - Load PGN works with no moves
+    {
+      pgn: [
+        '[Event "UrTurn Game Jam"]',
+        '[Round "6"]',
+        '[White "Kevo1ution"]',
+        '[Black "Billy Bob Jr."]',
+        '',
+      ],
+      expect: true
+    },
     {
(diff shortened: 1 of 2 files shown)


src/chess.ts
```diff
@@ -1655,7 +1659,7 @@ export class Chess {
     let moves = ms.trim().split(new RegExp(/\s+/))
 
     /* delete empty entries */
-    moves = moves.join(',').replace(/,,+/g, ',').split(',')
+    moves = moves.filter(move => move !== '')
```
kevo1ution, line 1662: The current method did not handle the edge case where `moves = ['']`:
```js
moves = [''].join(',').replace(/,,+/g, ',').split(',');
console.log(moves);
// output: ['']
```

The edge case happens when `ms` is empty (in the case where there are no moves):
```js
let ms = '';
console.log(ms.trim().split(new RegExp(/\s+/)));
// output: ['']
```

The solution I went with here is to just use the `.filter` method to remove any empty string from the array.

__tests__/load-pgn.test.ts
```diff
@@ -29,6 +29,21 @@ describe('Load PGN', () => {
...
       expect: true,
     },
+    // GitHub Issue #362 - Load PGN works with no moves
+    {
+      pgn: [
+        '[Event "UrTurn Game Jam"]',
+        '[Round "6"]',
+        '[White "Kevo1ution"]',
+        '[Black "Billy Bob Jr."]',
+        '', // extra newlineChar is expected because pgn() function adds it
```
kevo1ution, line 39: I left this here because there is a behavior with `pgn()` where it outputs the pgn headers and an extra `newlineChar`. I didn't want to change the behavior of the pgn() function in this PR - It's not clear whether we should or shouldn't change the `pgn()` fn, but either way that change should be in a separate PR.

kevo1ution:
Merry Christmas @jhlywa ! This is ready 🎄 🎅 - looking forward to the review

__tests__/load-pgn.test.ts
jhlywa: Agree with the separate PR.  Would it be simple enough to just make `.pgn()` eat any trailing `newlineChar`?

jhlywa:
Thanks @kevo1ution and Merry Christmas to you too!

PR #371, opened 2023-01-18
Fix type of Move.from and Move.to

MichaelTamm:
I think `Move.from` and `Move.to` should be of type `Square` and not `string`.

I run `npm run build`, `npm run lint`, and `npm run test` after the change -- all executed successfully.


jhlywa:
Good catch @MichaelTamm. Thank you!

PR #376, opened 2023-02-03
fix: parse minimal pgns

theborakompanioni:
This is currently WIP.

Goal of this PR is being able to parse minimal PGNs, e.g. starting positions (without any moves made yet).
Any feedback is highly appreciated. 
Q: What is the most minimal valid PGN?
From what it seems, an empty string `''` is a valid PGN, as mentioned in [Section 8 of the spec](https://github.com/mliebelt/pgn-spec-commented/blob/main/pgn-specification.md#8-parsing-games), hence it should be supported.
> "An empty file is a valid, although somewhat uninformative, PGN database"

The second minimum input would be just a Termination Marker, e.g. simply `*`.

This is for example how chess.com exports the starting position:
```pgn
[Event "?"]
[Site "?"]
[Date "????.??.??"]
[Round "?"]
[White "?"]
[Black "?"]
[Result "*"]

*
```

In all the cases described above, the `loadPgn` method throws an error:
```sh
Invalid move:

      1220 |     if (!moveObj) {
      1221 |       if (typeof move === 'string') {
    > 1222 |         throw new Error(`Invalid move: ${move}`)
           |               ^
      1223 |       } else {
      1224 |         throw new Error(`Invalid move: ${JSON.stringify(move)}`)
      1225 |       }
```


neofight78:
The cases you mention do work in the latest version of the library.

PR #386, opened 2023-03-23
Fixed moves() when there are no kings.

### LINKED ISSUE #218: king_attacked(color) returns true without king on board
Test:
```
chess.clear();
chess.put({type: chess.QUEEN, color: chess.BLACK}, 'd4');
chess.put({type: chess.KNIGHT, color: chess.WHITE}, 'a2');
console.log('strictly legal moves: ' + chess.moves({square: 'a2'}));
console.log('relaxed legal moves: ' + chess.moves({legal: false, square: 'a2'}));
```

the first console output will print `[]`, while the second output prints `[Nb4,Nc3,Nc1]`.
because the `king_attacked(color)` function returns true under some circumstances (e.g. the above test), even though no king is present.

Possible fix: check for a legal square in the `attacked(color, square)` function.
E.g. insert `if (square == -1) return false;` as the first line in the function. This would make sense, since `attacked(color, square)` should not return true for invalid squares anyways.

Note: the `square `parameter resolves to -1 in the above example because the kings array contains `[-1, -1]` if no kings are present.

neofight78:
fixes #218

### DIFF
--- a/__tests__/moves.test.ts
+++ b/__tests__/moves.test.ts
@@ -258,3 +258,13 @@ test('moves - square and piece', () => {
     split(moves)
   )
 })
+
+test('moves - no kings (starting position minus kings)', () => {
+  const noKings = new Chess();
+  noKings.remove("e1");
+  noKings.remove("e8");
+
+  const kings = new Chess();
+
+  expect(noKings.moves().filter(m => m !== "Qe1")).toEqual(kings.moves());
+})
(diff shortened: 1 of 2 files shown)


jhlywa:
Thanks again @neofight78! Can you verify that `inCheck`, `inCheckmate`, and `inStalemate` return `false` when the king is missing for the current side to move?

neofight78:
@jhlywa Good shout, inCheck fixed and tests added.

jhlywa:
Again, thank you!

PR #387, opened 2023-03-24
Ensure put/remove functions update castling rights

neofight78:
@jhlywa Once you are happy with this PR, I'll create a second one to ensure the ep square is updated also.

### DIFF
--- a/src/chess.ts
+++ b/src/chess.ts
@@ -49,6 +49,10 @@ export type Square =
     'a2' | 'b2' | 'c2' | 'd2' | 'e2' | 'f2' | 'g2' | 'h2' |
     'a1' | 'b1' | 'c1' | 'd1' | 'e1' | 'f1' | 'g1' | 'h1'
 
+export const KING_SIDE = 'k'
+export const QUEEN_SIDE = 'q'
+export type Side = 'k' | 'q'
+
 export const DEFAULT_POSITION =
   'rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1'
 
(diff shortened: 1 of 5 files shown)


jhlywa:
Thanks for the PR @neofight78. Is there anyway to setup a position where the king/rook are on the initial squares but castling rights are lost?  I'm wondering if it would be better to make an explicit `setCastlingRights` (or something similar) method?  We can certainly automatically remove rights if the kings/rooks aren't on their initial squares.

Thoughts?

neofight78:
@jhlywa It's possible either by using a fen, or starting with a position where the king/rook isn't in place and then updating it. But that certainly isn't ergonomic in all cases. So it certainly would make sense to add methods to set and get the castling rights directly. If you want to go for that, then I can certainly add those methods to this PR or a second one (which ever you prefer). In any case I think the code here for the automatic updating should be ok as is.

neofight78:
@jhlywa New methods added. Chose the public API that made the most sense to me, but let me know what you think.

jhlywa:
Thanks @neofight78. I'm finally getting a chance to take a look at this. What are your thoughts about making `setCastlingRight` and `getCastlingRight` take and return a "castling" object. I don't know, maybe something like this:

```js
// pluralized function names in this version

chess.setCastlingRights({ KING: true, QUEEN: false })

chess.getCastlingRights(WHITE)
// { 'k': true, 'q': false }
```

We could maybe reuse the existing `KING` and `QUEEN` exports instead of having to introduce `KING_SIDE/QUEEN_SIDE` exports? Maybe that's a bit more ergonomic, I dunno. API design is always tricky.

Any thoughts?

src/chess.ts
```diff
@@ -766,11 +777,34 @@ export class Chess {
...
     }
 
+    this._updateCastlingRights()
+
     this._updateSetup(this.fen())
 
     return piece
   }
 
+  _updateCastlingRights() {
```
jhlywa, line 787: Can you use the piece consts in place if the `'r'` and `'k'` literals here? Thanks!

jhlywa:
Looks great! Thanks again!

Issue #298, opened 2021-09-24
Ability to remove a header


justingolden21:
Let's say the players' names are set, but they remove a name. It would be nice to be able to remove a header (or all headers) rather than having to set it to an empty string. (Is it possible to set all headers at once to an empty object to attain this?)

PR #388, opened 2023-03-26
Added function removeHeader() to chess class

angarc:
Addresses #298


jhlywa:
Merged. Thanks for the PR @angarc!

PR #399, opened 2023-04-23
Make moveNumber accessible

neofight78:
It seems to me that the moveNumber should be accessible in the same way as the current turn. Being part of FEN it can't really be considered a private implementation detail.


jhlywa:
Good catch and thanks!

PR #412, opened 2023-05-27
fix: add check for pawn promotion in validateFen

Manukyanq:
If chess position has pawn on the last row, `chess.js` doesn't detect it as invalid fen, and the game continues as usual.

On the record below I've inited new game with fen where there is a pawn on the last row, (I'm playing as white, and blacks are playing random move from `game.moves()`
```
  const chess = new Chess("3P4/1k2K3/8/8/8/8/8/8 w - - 0 1")
```

https://github.com/jhlywa/chess.js/assets/42891870/e65c2315-7c23-46b1-a557-448cb9a7560b

### DIFF
--- a/__tests__/validate_fen.test.ts
+++ b/__tests__/validate_fen.test.ts
@@ -27,6 +27,16 @@ test.each([
     ok: false,
   },
   {
+    // white pawn on 8th row
+    fen: '3P4/1k2K3/8/8/8/8/8/8 w - - 0 1',
+    ok: false,
+  },
+  {
+    // black pawn on 1st row
+    fen: '8/8/8/8/8/8/1k3K2/3p4 w - - 0 1',
+    ok: false,
+  },
+  {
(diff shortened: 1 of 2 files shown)


Manukyanq:
Currently I'm checking only "promotion rows" considering pawn's color (we shouldn't have white pawns on the 8th row and black pawns on the first row).
 In classic chess rules we can not have any pawns on the 1st and 8th row, no matter its color, but I assumed that in some chess variants it can be ok to have white pawn on the first row(or black on the 8th).

Should I make the validation more strict based on classic chess rules or not?

P.S [lichess board editor](https://lichess.org/editor) considers any pawn on the 1st or 8th row invalid position

jhlywa:
Thanks for the PR @Manukyanq. Go ahead and make the rules more strict - any pawn on the 1/8th rank should be considered invalid.

jhlywa:
Accepted and thanks!

PR #414, opened 2023-06-04
non-strict FEN validation parameter

### LINKED ISSUE #413: can't setup initial fen without king
Hi I need a slight parameter to setup the board without the king, for custom boards purposes, how can I achieve that?

```Unhandled Runtime Error
Error: Invalid FEN: missing white king```

snk-js:
this closes #413 

image

### DIFF
--- a/__tests__/validate_fen.test.ts
+++ b/__tests__/validate_fen.test.ts
@@ -27,6 +27,16 @@ test.each([
     ok: false,
   },
   {
+    // white pawn on 8th row
+    fen: '3P4/1k2K3/8/8/8/8/8/8 w - - 0 1',
+    ok: false,
+  },
+  {
+    // black pawn on 1st row
+    fen: '8/8/8/8/8/8/1k3K2/3p4 w - - 0 1',
+    ok: false,
+  },
+  {
     fen: 'rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNRw KQkq - 0 1',
     ok: false,
   },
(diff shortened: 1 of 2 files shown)


jhlywa:
Hi and thanks @snk-js. This change is a bit too drastic and differs from the current chess.js conventions. Some methods currently use the optional `{strict: boolean()}` to allow illegal or non-standard behavior. My thinking is that we would add this optional parameter to the `.load()` method.  Unlike the other uses of strict, the default value for `.load` should be `{strict: true}`.

I don't want to extend this to the constructor at this stage, as it may imply enabling/disabling the strict/permissive parser across the object.

I'm really cautious about introducing large changes at this stage in the 1.0.0 release cycle, but I'm willing to accept a PR for the relatively minor modification to `load`.

Thanks again for the contribution.

snk-js:
> Hi and thanks @snk-js. This change is a bit too drastic and differs from the current chess.js conventions. Some methods currently use the optional `{strict: boolean()}` to allow illegal or non-standard behavior. My thinking is that we would add this optional parameter to the `.load()` method. Unlike the other uses of strict, the default value for `.load` should be `{strict: true}`.
> 
> I don't want to extend this to the constructor at this stage, as it may imply enabling/disabling the strict/permissive parser across the object.
> 
> I'm really cautious about introducing large changes at this stage in the 1.0.0 release cycle, but I'm willing to accept a PR for the relatively minor modification to `load`.
> 
> Thanks again for the contribution.

I changed as follows, I had already do that way, and it was already optionally too. I do not  think this is a drastic. Because we are dealing bypassing validations per se, it doesn't changed much from the original but I hope it's acceptable, many of the changes in this PR was formatting code. Let me know if this attends more briefly.

PR #416, opened 2023-06-10
fix readme shield links

JamJomJim:
The readme badges currently don't link to anything (they just link to the the image's URL). I've updated these to link to the relevant URL.


jhlywa:
Thanks @JamJomJim!

Issue #144, opened 2017-03-01
Threefold repetition checking


imilo:
**in_threefold_repetition()** returns **true** when the same position has repeated 3+ times, but if after that I make absolutely another moves, the opponent too, **in_threefold_repetition()** still returns **true**.

I'm not sure that this behavior is by the chess rules. I think the result should be **false** when the position has changed

jhlywa:
Hi @imilo.  Thanks for the report.  I think you're correct.  Using FIDE rule 9.2 as a reference:

```
The game is drawn, upon a correct claim by a player having the move, when the same 
position for at least the third time (not necessarily by a repetition of moves):

- is about to appear, if he first writes his move, which cannot be changed, on his
  scoresheet and declares to the arbiter his intention to make this move, or

- has just appeared, and the player claiming the draw has the move.
```

So maybe we should change the code so `in_threefold_repetition()` returns true if and only if the **current** position has been repeated three times?

imilo:
@jhlywa , if that question was for me, that's a tricky question...

For example, if I make a position that has appeared 3+ times -> `in_threefold_repetition()`  will return True a that moment. But my opponent can make his new unique move very quickly -> position will be changed -> `in_threefold_repetition()`  will return False for me. So I will lose my chance to claim for a draw.

`chess.js` should handle this somehow. JavaScript could remember that I can claim for a draw until my next move, so we don't care about the opponent's move. These are my thoughts, may be you have a better solution

dsjoerg:
@imilo you make a good point.  I think we should change `in_threefold_repetition()` to return true _only_ if the _current_position_ has appeared for at least the third time.  That way we match FIDE rules.   @jhlywa if you're OK with this I'll submit a PR for your approval.

@imilo I like your suggestion that Javascript can remember that you can claim for a draw until your next move.  That's not the FIDE rule, but you can make it that way for your own chess site even if chess.js is strictly following the FIDE rules.  You could do it like this:

```
if (players_turn) {
  if (game.in_threefold_repetition()) {
    // show "Claim Draw" button to the player
  } else {
    // check and see if our last move created a threefold repetition so that we could have claimed a draw.
    var move = game.pop()   // undo the opponent's move
    if (game.in_threefold_repetition()) {
      // show "Claim Draw" button to the player
    }
    game.make_move(move)
  }
}
```

ajax333221:
[...]

PR #417, opened 2023-06-14
Changes to `isDraw` and `isGameOver` (to fix #112) and fixes for games that include `put`/`remove`

### LINKED ISSUE #112: Drawn positions not determined correctly (technically)
[Laws of Chess (Draws)](https://en.wikipedia.org/wiki/Rules_of_chess#Draws)

The Laws state that at least one of the players must claim the draw whenever three-fold repetition occurs or the fifty-move rule may be invoked. In other words, a draw is not automatic whenever one of these conditions is met. This means that the game.game_over() and game.is_draw() methods cannot be relied on in serious applications. Furthermore, there is no way to determine through the API how many half moves have elapsed since the previous capture or pawn move.

My first inclination is to expose three new methods that abide by the strict interpretation of the Laws: game_over_strict(), in_draw_strict() and in_fifty_move_rule(). They would respect the following rules. (I omit the game object for brevity:)

game_over_strict() === in_checkmate() || in_draw_strict()
in_draw_strict() === in_stalemate() || insufficient_material()
in_fifty_move_rule() === half_moves >= 100

I would be happy to hear any feedback.

gavin-lb:
# Changes 
 - Fixed issue #144 (longstanding bug in `isThreefoldRepetition`)
 - Added API methods: `isFivefoldRepetition`, `isFiftyMoveRule`, `isSeventyFiveMoveRule`
 - Changed `isDraw` and `isGameOver` methods to accept an optional `strict` parameter (defaults to false to preserve backwards compatibility). When true, they will strictly adhere to the lastest FIDE rules regarding drawn games (see [Article 9 of the Laws of Chess](https://handbook.fide.com/chapter/E012023)) fixing issue #112 
 - Added tests for the new functionality
 - Updated docs

### DIFF
--- a/src/chess.ts
+++ b/src/chess.ts
@@ -1642,17 +1642,17 @@ export class Chess {
     return old
   }
   
-  private _redoMove({historyType, move}): void {
+  private _redoMove({historyType, move}: HistoryType): void {
     // Used to redo moves undone with _undoMove (not with undo)
     switch (historyType) {
       case 'move':
-        this._makeMove(move)
+        this._makeMove(move as InternalMove)
         break
       case 'put':
-        this.put(move.piece, move.square)
+        this.put(move.piece as Piece, (move as Put).square)
         break
       case 'remove':
-        this.remove(move.square)
+        this.remove((move as Remove).square)
         break
     }
   }
@@ -1700,7 +1700,7 @@ export class Chess {
     // pop all of history onto reversed_history
     const reversedHistory: History[] = []
     while (this._history.length > 0) {
-      reversedHistory.push(this._undoMove())
+      reversedHistory.push(this._undoMove()!)
     }
 
     const moves = []
@@ -1713,7 +1713,7 @@ export class Chess {
 
     // build the list of moves.  a move_string looks like: "3. e3 e6"
     while (reversedHistory.length > 0) {
-      const { historyType, move } = reversedHistory.pop()
+      const { historyType, move } = reversedHistory.pop()!
 
       if (historyType === 'move') {
         const moveEvent = move as InternalMove
@@ -2381,7 +2381,7 @@ export class Chess {
   history({ verbose, onlyMoves }: { verbose: true, onlyMoves: false }): HistoryMove[]
   history({ verbose, onlyMoves }: { verbose: false, onlyMoves: false }): never
   history({ verbose, onlyMoves }: { verbose: boolean, onlyMoves: boolean }):  string[] | Move[] | HistoryMove[]
-  history({ verbose = false, onlyMoves = true }: { verbose?: boolean, onlyMoves?: boolean } = {}) {
+  history({ verbose = false, onlyMoves = true }: { verbose?: boolean, onlyMoves?: boolean } = {}):  (string | Move | HistoryMove)[] {
     /*
      * TODO: Why does history need to replay the entire game to get the history? We are already
      * storing the game history in _history. It seems the only reason we are replaying the entire 
@@ -2398,11 +2398,11 @@ export class Chess {
     const moveHistory = []
 
     while (this._history.length > 0) {
-      reversedHistory.push(this._undoMove())
+      reversedHistory.push(this._undoMove() as History)
     }
 
     while (reversedHistory.length > 0) {
-      const { historyType, move } = reversedHistory.pop()
+      const { historyType, move } = reversedHistory.pop() as History
 
       if (onlyMoves && historyType === 'move') {
         if (verbose) {
@@ -2414,7 +2414,7 @@ export class Chess {
         if (historyType === 'move') {
           moveHistory.push({ historyType, move: this._makePretty(move as InternalMove) })
         } else {
-          moveHistory.push({ historyType, move })
+          moveHistory.push({ historyType, move: move as Put | Remove })
         }
       }
 


jhlywa:
Thanks for the PR @gavin-lb. I'll do my best to review these changes (and familiarize myself with the latest rule updates) sometime this weekend. Since we're technically pre-1.0.0, I'm comfortable breaking backwards compatibility for some of these edge-case draws - meaning we might not need to include the strict parameter on `isDraw` and `isGameOver`.

gavin-lb:
Well I think it's probably more common for people to want auto claiming of draws on threefold repetition or 50-move rule anyway, so still makes sense for that to be the default (for example, I'm pretty sure Lichess does it that way). But for those that want to implement a stricter adherence to the FIDE rules as written, the option and bug fix are necessary.

gavin-lb:
Added a couple optimisations to this PR, but let me know if you would rather they be a separate PR instead.

gavin-lb:
Fixed a bug with the new `_getRepetitionCounts` implementation: it wasn't adding positions to the record when loading from PGN (sorry ignore the commit name when I said FEN, I meant PGN).

However, it is a little unclear to me what the behaviour should be with regards to repetitions and `put`/`remove`. There is no precedence set since the previous implementation didn't consider these pieces at all (since they are not added to `_history`) - which I suppose could have been considered a bug. But it leaves unanswered how the new implementation should handle these cases. At present, `_positionCount` will not be updated on the turn in which pieces are put or removed from the board, but all subsequent turns will include the put/removed pieces in the position. Although, again, since these pieces are not added to the history, things will be messed up if a piece is put/removed from the board and then moves are undone (the piece will remain on the board and the previous positions will not be correctly removed from `_positionCount` and indeed previous moves might now be illegal with the addition/removal of the piece). Any thought on how to handle these cases?

gavin-lb:
Using  `put`/`remove` already break a lot of functionality. Because they are not in the `_history` but still affect `_board`: `pgn`, `history` and `undo` methods all potentially produce non-sense output when a game includes a `put`/`remove` (particularly regarding SAN).  

I think the only obvious way to amend the aforementioned problem with `_positionCounts` when undoing, as well as the existing problems with `pgn`, `history` and `undo` in general, is to **keep track of  `put`/`remove` calls by adding them to `_history`**. 

This will allow `undo` to properly undo the `put`/`remove`, therefore allowing `_positionCounts` to properly remove previous positions from the record and also allowing `history` to generate accurate output. The `pgn` method will still not really be able to generate proper output because (as far as I can tell) there is no way to specify arbitrarily adding/removing pieces from the board in the PGN specification. However, it would at least allow the note of additions/removals to the board via a PGN comment and, whilst the output would still not be a valid PGN as per the specification, it would at least produce something that makes more sense (all the SAN moves would be accurate if accounting for the added/removed piece). 

This change would require some alterations across the whole library and would technically affect backwards compatibility. However, it would only affect those users that used `put`/`remove` as well as `undo`/`history`/`pgn` and given the currently broken state of those methods with `put`/`remove`, I don't expect anyone is doing this. The backwards compatibility of users that don't use  `put`/`remove` will be unaffected.

jhlywa:
> Well I think it's probably more common for people to want auto claiming of draws on threefold repetition or 50-move rule anyway, so still makes sense for that to be the default (for example, I'm pretty sure Lichess does it that way). But for those that want to implement a stricter adherence to the FIDE rules as written, the option and bug fix are necessary.

Yep, you're right. I remember now. I'm in agreement with you here. 

As for the API changes to `isDraw` and `isGameOver`, would it be better to modify the `Chess` constructor to take an optional `strictFideAdherence` flag (or something similar)? I don't know if I have a strong preference either way, but I'd like to explore all options when modifying the API.

> Added a couple optimisations to this PR, but let me know if you would rather they be a separate PR instead.

@gavin-lb Let's split those into a new PR and I'll merge them today.

> However, it is a little unclear to me what the behaviour should be with regards to repetitions and put/remove.

My opinion is that `.put()`, `.remove()`, and `.clear()` should be considered destructive operations and any call should reset history, move counters, positions counts, etc.. These methods don't do this right now - my fault - but they should. I honestly haven't given much thought to the use-cases involving these methods, so I could be totally wrong here.

gavin-lb:
> As for the API changes to isDraw and isGameOver, would it be better to modify the Chess constructor to take an optional strictFideAdherence flag (or something similar)? I don't know if I have a strong preference either way, but I'd like to explore all options when modifying the API.

That's certainly another viable approach. I'm not sure which is preferable really. I suppose passing a flag to the constructor would be less flexible but would promote greater consistency. 

> Let's split those into a new PR and I'll merge them today.

I've submitted a new PR with only the optimisations (although, the bug fix is a dependency of the `isThreefoldRepetition` optimisation, so I've also included that). 

I will repurpose this PR for changes to the API methods `isDraw` and `isGameOver` and how to potentially solve the issues with `put`/`remove`. 

> My opinion is that .put(), .remove(), and .clear() should be considered destructive operations and any call should reset history, move counters, positions counts, etc.. These methods don't do this right now - my fault - but they should. I honestly haven't given much thought to the use-cases involving these methods, so I could be totally wrong here.

Ah that would be a good way to deal with it as it would create no ambiguity at all. But I suppose it is less powerful since clearing the history removes a lot of functionality from the user. 

I will push some commits to this PR that show how the solution of keeping track of `put`/`remove` calls by adding them to `_history` might be implemented.

gavin-lb:
Latest commits show how `_history` could be updated to include `put`/`remove`s, including updates to `pgn` and `loadPgn` methods to they can output and read from pgns with `put`/`remove`s commented like this: 

    '1. e4 {Chess.js: w p put on e2} {Chess.js: w p removed from d2} e5'

ANDREA-1720:
Hi... before finding this PR, I've created a smaller one with similiar changes to the draw handling

#553

PR #418, opened 2023-06-18
Fix #144 and optimisations

### LINKED ISSUE #144: Threefold repetition checking

gavin-lb:
As discussed in [the other PR](https://github.com/jhlywa/chess.js/pull/417), I'm splitting the optimisations into their own PR. 

Since the bug fix is a dependency of the `isThreefoldRepetition` optimisation, I've included it in this PR.

### DIFF
--- a/__tests__/is-threefold-repetition.test.ts
+++ b/__tests__/is-threefold-repetition.test.ts
@@ -12,6 +12,8 @@ test('isThreefoldRepetition', () => {
     chess.move(move)
   })
   expect(chess.isThreefoldRepetition()).toBe(true)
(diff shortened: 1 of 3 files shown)


ajax333221:
there is an optimization using the half-move clock, you can start counting the positions from newer to older and stop when encounter a 0 in there.

PR #419, opened 2023-06-20
Follow up to PR #418

gavin-lb:
Sorry, just realised I forgot to include some changes when splitting off the last PR from the other one:

 - removed the unnecessary method calls from `_put`
 - removed unnecessary `.fen` call from `load`


gavin-lb:
Actually, I'll branch this from the current main branch rather than from my old branch

PR #426, opened 2023-07-15
fix: changing piece on the kings square with chess.put()

Manukyanq:
Hello! I want to make a chessboard, where users can manually set the position they need and  the`chess.put()` and `chess.remove()` methods are superuseful for me. I found a tiny bug, which I can make work with small hack, but anyway it is confusing and will be nice to fix it.

Lets say I want manually swap white king's and white queen's positions.

```
const chess = new Chess()

const status1 = chess.put({ type: 'q', color: 'w' }, 'e1') // true
const status2 = chess.put({ type: 'k', color: 'w' }, 'd1') // false

```

but if I remove white king and only after then put queen and king it works without errors

```
const chess = new Chess()

chess.remove('e1')
const status1 = chess.put({ type: 'q', color: 'w' }, 'e1') // true
const status2 = chess.put({ type: 'k', color: 'w' }, 'd1') // true

```

I can provide more details if it is required. Thank you

### DIFF
--- a/src/chess.ts
+++ b/src/chess.ts
@@ -808,6 +808,7 @@ export class Chess {
 
     const currentPieceOnSquare = this._board[sq]
 
+    // if one of the kings will be replaced by the piece from args, set the `_kings` respective entry to `EMPTY`
     if (currentPieceOnSquare && currentPieceOnSquare.type === KING) {
       this._kings[currentPieceOnSquare.color] = EMPTY
     }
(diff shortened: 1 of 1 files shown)


jhlywa:
Hi @Manukyanq, thanks for the PR!

I think I'm tracking the issue.  Essentially, if you overwrite the position of the king with any other piece, we need to set the `_kings` entry (for the overwritten king color) to `EMPTY`.  Could you add a comment to that section just to highlight the what's happening?  A one-liner is fine.

Thanks and good find!

jhlywa:
Merged. Thanks again!

PR #437, opened 2023-09-25
add default separator param to ascii function

kwkr:


neofight78:
@jhlywa An old PR, but idea looks reasonable, if we accept a newline character for pgn()? (We'd want to rename the parameter on this PR to match).

neofight78:
@jhlywa @neofight78 An old PR, minor but worthwihle I think. I changed the parameter to newline to match the parameter for pgn(), also updated the docs to match. Should be ready to merge now.

PR #451, opened 2024-03-18
#319: add null move functionality

ppeloton:
Implemented null move logic for SAN '--' and added a simple test case. Please review, I am open for discussions!

### DIFF
--- a/src/chess.ts
+++ b/src/chess.ts
@@ -101,6 +101,7 @@ const FLAGS: Record<string, string> = {
   PROMOTION: 'p',
   KSIDE_CASTLE: 'k',
   QSIDE_CASTLE: 'q',
+  NULL_MOVE: 'nm',
 }
 
 // prettier-ignore
@@ -123,6 +124,7 @@ const BITS: Record<string, number> = {
   PROMOTION: 16,
   KSIDE_CASTLE: 32,
   QSIDE_CASTLE: 64,
+  NULL_MOVE: 128,
 }
 
 /*
@@ -264,6 +266,8 @@ const ROOKS = {
 const SECOND_RANK = { b: RANK_7, w: RANK_2 }
 
 const TERMINATION_MARKERS = ['1-0', '0-1', '1/2-1/2', '*']
+
+const SAN_NULLMOVE = '--'
 
 // Extracts the zero-based rank of an 0x88 square.
 function rank(square: number): number {
@@ -1405,6 +1409,18 @@ export class Chess {
     const them = swapColor(us)
     this._push(move)
 
+    if (move.flags & BITS.NULL_MOVE){
+      if (us === BLACK) {
+        this._moveNumber++
+      }
+  
+      this._turn = them
+
+      this._epSquare = EMPTY
+
+      return
+    }
+
     this._board[move.to] = this._board[move.from]
     delete this._board[move.from]
 
@@ -1523,6 +1539,10 @@ export class Chess {
 
     const us = this._turn
     const them = swapColor(us)
+
+    if (move.flags & (BITS.NULL_MOVE)) {
+      return move
+    }
 
     this._board[move.from] = this._board[move.to]
     this._board[move.from].type = move.piece // to undo any promotions
@@ -1961,6 +1981,8 @@ export class Chess {
       output = 'O-O'
     } else if (move.flags & BITS.QSIDE_CASTLE) {
       output = 'O-O-O'
+    } else if (move.flags & BITS.NULL_MOVE){
+      output = SAN_NULLMOVE
     } else {
       if (move.piece !== PAWN) {
         const disambiguator = getDisambiguator(move, moves)
@@ -1981,15 +2003,17 @@ export class Chess {
       }
     }
 
-    this._makeMove(move)
-    if (this.isCheck()) {
-      if (this.isCheckmate()) {
-        output += '#'
-      } else {
-        output += '+'
+    if (!(move.flags & BITS.NULL_MOVE)){
+      this._makeMove(move)
+      if (this.isCheck()) {
+        if (this.isCheckmate()) {
+          output += '#'
+        } else {
+          output += '+'
+        }
       }
+      this._undoMove()
     }
-    this._undoMove()
 
     return output
   }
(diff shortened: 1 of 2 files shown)


jhlywa:
Thanks for the PR @ppeloton. 

This is a good start on the null move logic. A couple of random questions that I haven't really thought through yet:

- Are null moves supported in the PGN logic (both loading and exporting)? 
- Should null moves be entered with a dedicated function instead of the `.move()` function? I think there's a lot to think about here. A dedicated function would keep `.move()` consistent with the results of `.moves()`. It also side-steps the issue of not being able to enter a null move when using the verbose move format. A lot of people have been asking for `setSideToMove` function. Could this be considered a null move? Could we combine the two?
- Should a null move entered at move 1 (`this._history.length === 0`) be omitted from history? Omitting it would require modifying the starting FEN and updating the `Setup` header. This relates to the previous question about `setSideToMove`.
- Can null moves be made in succession?
- Can null moves be made when the side to move is in check?

I'd like to let this PR soak for a bit to let us (and others) ponder some of these questions. Thanks again for submitting!

src/chess.ts
```diff
@@ -101,6 +101,7 @@ const FLAGS: Record<string, string> = {
   PROMOTION: 'p',
   KSIDE_CASTLE: 'k',
   QSIDE_CASTLE: 'q',
+  NULL_MOVE: 'nm',
```
jhlywa, line 104: The FLAGS.NULL_MOVE value should be a single character. As written, the `n` in `nm` collides with FLAGS.NORMAL. Maybe use '-'? 

The whole flag system is an artifact of some C code I wrote years ago, and is kinda kludgy.

src/chess.ts
```diff
@@ -1961,6 +1981,8 @@ export class Chess {
       output = 'O-O'
     } else if (move.flags & BITS.QSIDE_CASTLE) {
       output = 'O-O-O'
+    } else if (move.flags & BITS.NULL_MOVE){
+      output = SAN_NULLMOVE
```
jhlywa, line 1985: Could you just return SAN_NULLMOVE here to skip the rest of this function?

src/chess.ts
ppeloton: I fixed this with the suggested value '-'

src/chess.ts
ppeloton: Good point, I fixed this.

ppeloton:
@jhlywa : Happy to contribute and thank you for your good feedback and questions. Generally, I think that null moves do not have one common standard, so if we want to implement this in the package, the developers have to make choices and document how null moves are handled in chess.js 

About your questions:
- Exporting/Importing in pgn: Just the SAN for a null move has no standard (see e.g. https://chess.stackexchange.com/questions/14072/san-for-nullmove), so exporting / importing of null moves is probably differently supported. I personally use Scid, which supports importing/exporting of null moves in pgn files (using SAN '--').

- a seperate method for null moves sounds good (e.g. `makeNullMove()`) sounds like a good way to implement it. Even more if there is a wish for `setSideToMove` functionality (if I understand it correctly, it basically corresponds to making a null move and giving the turn to the otherside).

- Rules relating null moves: I have not found a standard for this. But using Scid as a reference, the program allows multiple null moves in succession but no null moves when one is in check. I could try to find out if there are more rules in Scid related to null moves and we good take that as a starting point for this package.

I am happy to continue the discussion!

ppeloton:
Sorry, I forgot to run the prettify command. I hope now everything is fixed!

EarlyEdition:
> * a seperate method for null moves sounds good (e.g. `makeNullMove()`) sounds like a good way to implement it. Even more if there is a wish for `setSideToMove` functionality (if I understand it correctly, it basically corresponds to making a null move and giving the turn to the otherside).

`makeNullMove()` is useful and necessary for developing a chess engine in JavaScript. 
I hope that this PR will be merged as soon as possible.

neofight78:
Just a few thoughts:

1) I agree with @ppeloton that -- is probably the best / most widely used notation for null moves in PGNs.

2) I dislike having a dedicated makeNullMove() function, I think it would be a lot cleaner if a null move worked in the same way as any other move. That also leads to a simple answer to your other queries: i.e. it's always inserted into the history, you can have multiple null moves in a row, but it's not a legal move when in check (as you will still in check "after" the move). For the verbose move, we can give sensible values for some fields and others leave as undefined perhaps?

3) I think if null move is implemented, then that negates the need for set side to move. I'm not sure how you would change the side to move halfway through a game otherwise.

jhlywa:
I agree, I don't think null move entry needs its own API.  I think `move('--')` and `move(null)` should both be allowed.

I think `setTurn` (or `setSideToMove`) can be defined in terms of the null move, but should be it's own function. It would be more idempotent, generating 0 or 1 null moves depending on the original side to move.

src/chess.ts
```diff
@@ -1405,6 +1409,18 @@ export class Chess {
     const them = swapColor(us)
     this._push(move)
 
+    if (move.flags & BITS.NULL_MOVE) {
+      if (us === BLACK) {
+        this._moveNumber++
+      }
+
```
neofight78, line 1416: I think we need to update the half move clock too?

src/chess.ts
```diff
@@ -1981,15 +2003,17 @@ export class Chess {
       }
     }
 
-    this._makeMove(move)
-    if (this.isCheck()) {
-      if (this.isCheckmate()) {
-        output += '#'
-      } else {
-        output += '+'
+    if (!(move.flags & BITS.NULL_MOVE)) {
```
neofight78, line 2006: Is this check necessary? Don't we return early if it's a null move?

src/chess.ts
```diff
@@ -2202,9 +2238,13 @@ export class Chess {
...
-    this._makeMove(uglyMove)
-    move.after = this.fen()
-    this._undoMove()
+    if (!(uglyMove.flags & BITS.NULL_MOVE)) {
+      this._makeMove(uglyMove)
+      move.after = this.fen()
+      this._undoMove()
+    } else {
+      move.after = this.fen()
+    }
```
neofight78, line 2247: Not sure this change is needed, don't we need to make the null move and return the new fen with different color and move number etc?

neofight78:
@ppeloton Can you rebase and resolve any conflicts? See @jhlywa's latest comments, I think allow move(null) too and it's pretty much there. I've also added a few questions on the implementation, perhaps some tweaks are needed. Then the PR should be ready to merge. If you want bonus points a few more tests would be great.

ppeloton:
@neofight78 @jhlywa thanks for the comments. I will have a look at the comments and try to make the necessary changes in the next coming days.

src/chess.ts
ppeloton: Correct, I added this in the code and also adapted the test cases

src/chess.ts
ppeloton: Correct, I removed this part

ppeloton:
I made adaptions to the code: 

- code adaptions based on comments (redundant lines, half move counter)
- `move(null)` is now supported
- null move when in check is not allowed
- i added several test cases

neofight78:
Thanks @ppeloton all looks good! Build is failing on a formatting check though. Can you run: npm run format and push the changes?

neofight78:
Thanks @ppeloton !

@jhlywa This PR looks good to merge now unless you have any final comments.

jhlywa:
Merged.  Thanks for the PR @ppeloton!

PR #458, opened 2024-05-03
Fix minor error in chess.moves({ verbose: true }) output

TheLetterZed:
Looks like a typo or a minor copy paste error in the output of this part of the readme.


jhlywa:
Thank you @TheLetterZed!

PR #487, opened 2025-02-16
correct chess960 castling

sowson:
Hi @jhlywa, Thanks for a great project. I improved a bit. Now casting in 960 works! Tested at https://iChess.io Enjoy!


jhlywa:
Hi @sowson. Thanks for the PR!

For 960, I'd like to use a separate class for the 960 entry point (e.g. `const chess = new Chess960()`). That should give us good flexibility in supporting other variants in the future without polluting the main `Chess` class. So ... maybe extend from `Chess` and override methods where needed (not sure if this is possible or not). Also note that there are required PGN headers for 960 games.

That said, a huge red flag for me is the modification of perft and regression tests for the existing code. A perft counts the number of legal positions at a given depth from a starting position. If perft results ever change, it means the move generation algorithm is broken. We simply cannot let that happen.

See https://www.chessprogramming.org/Perft_Results and https://www.chessprogramming.org/Chess960_Perft_Results for more info.

jhlywa:
Castling queenside is a legal move for white in that position.

Further, the original castling rights of 'KQ' are accurate. From the Wikipedia on [FEN](https://en.wikipedia.org/wiki/Forsyth%E2%80%93Edwards_Notation) (emphasis is mine):

3.  Castling availability: If neither side has the ability to castle, this field uses the character "-". Otherwise, this field contains one or more letters: "K" if White can castle kingside, "Q" if White can castle queenside, "k" if Black can castle kingside, and "q" if Black can castle queenside. **A situation that temporarily prevents castling does not prevent the use of this notation.**

sowson:
You are right! :D And I learned a lot today! :-). I have the solution and have passed all the tests except one. Do not get me wrong, but it looks like this test is OK now. Chess is great! Thanks!

sowson:
There is one more thing I found ;-). Tests are ~5s now instead of ~7s long and there are 13 more to magic 404 number :D.
Your 1.0.0 check:
Screenshot 2025-02-16 at 16 48 18

Mine version check:
Screenshot 2025-02-16 at 16 48 45

Thanks! Please consider the contribution of this PR.

jhlywa:
I'm going to attempt to merge this into the 960 branch so we can collaborate on it for a bit. The 960 branch is outdated by 13 years, so the merge may fail dramatically. Stay tuned.

MaximMolchanov:
Guys pls consider parsing X-FEN as [suggested in the wiki](https://en.wikipedia.org/wiki/Forsyth%E2%80%93Edwards_Notation#FEN_adjustment_for_chess_variants_like_Chess960) and [as stockfish does](https://github.com/official-stockfish/Stockfish/blob/fa6c30af814fe91e6a6c2d1bcaa8d951e3724ae7/src/position.cpp#L232-L260)

sowson:
Hi @jhlywa, in the last commit, I added support for X-FEN to my fork (this PR) and some more tests. Is that now OK? :D.

jhlywa:
Hi @sowson. Good work! I merged this into the 960 branch for collaboration.

I know you'd like to see this merged into master quickly, but there's still more to do before it's ready. Off the top of my head:

  - `.load()` needs to support X-FEN castling rights
  - changes to `pgn`, `validateFen`, `getCastlingRights`, `setCastlingRights`
  - add a public `is960()` method
  - documentation (both code and README for all impacted public methods)
  - 960 perft tests from https://www.chessprogramming.org/Chess960_Perft_Results

I can help with some of these (time permitting). Thanks again and I'm excited about this PR!

PR #489, opened 2025-02-19
Chess 960 Improvements

sowson:
Hi @jhlywa, I'm working on that one. It now needs more description in the README.md file only. I am still looking into some issues with the Perf Test of Chess 960. With Depth 2, I have 4 different values, and on Depth 3, I have 21 different values. I'm not sure why. Thanks!

PR #493, opened 2025-02-28
Add Chess960 support.

deverac:
I do not know if Chess960 support is wanted, but I added it for my own amusement. If you don't want it, please feel free to reject or close this Pull Request.

To perform a castle operation, move the king onto the target rook.

The tests cover every possible castling combinations of one king and one rook (black/white/kingside/queenside).

If you want changes, please let me know.


jhlywa:
Hi @deverac and thanks for the PR!

There have been lots of request for 960 support lately, so this PR is definitely needed. @sowson has been been working on PR #489, but ran into a few issues with a few of the perft positions. A couple of questions:

1. I don't play 960, is moving the king on top of rook the standard mechanic for castling in 960? If the king ends up in the same position as classical chess, wouldn't e1-g1 or e1-c1 be the correct notation?

2. There are requests to support X-FEN notation specifically for 960 castling rights? Do you think the library should require X-FEN if the user wants to load a 960 position? The presence of X-FEN castling rights (in `load()`) could automatically put the library into 960 mode without the need for additional parameters.

3. Do you have any idea if the performance of the classical move generator is impacted by the 960 changes? The sort on line 1617 ( `const moves = this._moves().sort((a, b) => b.flags - a.flags))` has me a little concerned. I'm still looking for a good way to reliably benchmark perft performance.

This is a great start! Thank you very much for the PR!

sowson:
Well, I copied x-fen and perft tests from my PR to this one locally to check and... well...
Screenshot 2025-03-01 at 11 44 31

sowson:
@deverac, can you do the opposite check I mentioned above? Could you test my code against your tests?
Screenshot 2025-03-01 at 11 49 46

sowson:
@jhlywa I took all the tests, mine and the @deverac, and then I ran on my repo... I think it is pretty good. Thanks!

Screenshot 2025-03-01 at 11 52 33

sowson:
Hi, @deverac. I love that you did the performance tests. Can I took to my PR #489 ?? :).

sowson:
Hi, @deverac. Can you please check the test below?

`
  it('X-FEN O-O 960 28', () => {
    const chess = new Chess(
      'rnbkrbqn/p1pp1ppp/4p3/1p6/8/BPN3P1/P1PPPP1P/R2KRBQN w AEae - 2 9',
      { enable960: true}
    )
    chess.loadPgn('9. Bh3 c6 10. Qg2 g6')
    chess.move({ from: 'd1', to: 'g1' })
    expect(chess.moves().length).toBe(36);
    expect(chess.fen()).toEqual(
      'rnbkrbqn/p2p1p1p/2p1p1p1/1p6/8/BPN3PB/P1PPPPQP/R4RKN b ea - 1 11',
    )
  })
`

sowson:
And this one:

`
  it('X-FEN O-O 960 29', () => {
    const chess = new Chess(
      'rnbkrbqn/p2p1p1p/2p1p1p1/1p6/8/BPN3PB/P1PPPPQP/R2KR2N w EAea - 0 11',
      { enable960: true}
    )
    expect(chess.moves().length).toBe(36);
    chess.move({ from: 'd1', to: 'g1' })
    expect(chess.fen()).toEqual(
      'rnbkrbqn/p2p1p1p/2p1p1p1/1p6/8/BPN3PB/P1PPPPQP/R4RKN b ea - 1 11',
    )
  })
`

deverac:
I apologize for the following wall of text, but I believe the answers will require a bit of explanation.

> is moving the king on top of rook the standard mechanic for castling in 960?

In Standard chess, an opponent can deduce the intent of a player simply by observing how the player moves the pieces. When it comes to castling, moving the king to its final destination square requires the king to move two spaces. (If the king moves more than one space, than it is obvious that the player is intending to castle.) Castling is trickier in Chess960. In Chess960 there are some positions where castling requires the king to move just one square. In such positions, it is impossible to know the intent of the player that is moving the king simply by observing how the player moves the pieces. Once the king has begun to move, the *only* difference between castling and simply moving the king a single space is the intent of the player that is moving the king.
(In other Chess960 positions, the king doesn't move at all when castling.)

[FIDE](http://rcc.fide.com/guidelinesii/) has this to say about castling in Chess960: "When castling on a physical board with a human player, it is recommended that the king be moved outside the playing surface next to his/her final position, the rook then be moved from its starting position to its final position, and then the king be placed on his final square."
and "To avoid any misunderstanding, it may be useful to state 'I am about to castle' before castling."

Implementing this behavior for a virtual board could be done, but would be complicated. To simplify the implementation on virtual boards, one mechanism that has been devised is to move the king onto the rook that it wishes to castle with. This mechanic that is really only needed in order to disambiguate a single move vs. a castling move. In most positions, simply moving the king
to its final destination square requires the king to move more than one space, so the intent
to castle is clear.

I don't think there is a 'standard' way for how to accomplish a castling move in Chess960, but
Stockfish has adopted this mechanic.

Instead of moving the king onto the rook, various other mechanics could be used.

The king-onto-rook move is not needed for notating games since castling is indicated by 'O-O' and 'O-O-O'.


> If the king ends up in the same position as classical chess, wouldn't e1-g1 or e1-c1 be the correct notation?

Yes, those notations are correct, but only if the king starts on 'e1'. In Chess960, the white king can start on any square in row 1 except for 'a1' and 'h1'. The black king can start on any square in row 8 except for 'a8' and 'h8'. Exactly one rook is positioned on either side of the king may appear on any square. So, if the white king starts on 'f1', the move 'f1-g1' is ambiguous. Is the king simply moving to 'g1' (in which case, the rook should not move), or is it a kingside castling-move (in which case, the rook should move). Moving the king 'onto' the rook is one way of indicating a castling-move.
 

> Do you think the library should require X-FEN if the user wants to load a 960 position?


In order to load a 960 position, the library will require a *partion* of X-FEN. It is not necessary for all of X-FEN to be supported in order to load a 960 position. I could easily be wrong, but I suspect that requests to support X-FEN are actually requests to support Chess960 (which  requires implementing a portion of X-FEN).

It seems that an actual X-FEN specification does not exist. Wikipedia has a link to the specification, but it is a dead link and apparently stopped working in 2020. I am unable to find an X-FEN specification, so the only thing I have to go on is various examples on the web.

Per [wikipedia](https://en.wikipedia.org/wiki/X-FEN#X-FEN_example), X-FEN supports other variants of chess and adds additional pieces ('A', 'C'), a larger board (8x10), additional castling flags ('s', 'm') and a stricter en-passant field. X-FEN also allows the ability to encode castling rights for Fischer Random chess (a.k.a. Chess960) which cannot be accurately encoded using standard FEN. There may be other differences that are not mentioned.

> The presence of X-FEN castling rights (in load()) could automatically put the library into 960 mode without the need for additional parameters.

I believe the short answer is No.

It seems that X-FEN allows the use of 'KQkq' castling flags in a FEN-string that describes a Chess960 position. I suspect this was done for backward compatibility in order to make it easy to generate and supply a Chess960 starting position to engines (and people) in order to start playing a Chess960 game. The 'KQkq' castling flags in the FEN-string were simply interpreted to mean the outermost rooks. This works perfectly fine when starting a new Chess960 game. Using 'KQkq' castling flags to encode which rooks have castling rights works for almost any arbitrary Chess960 board; it just can't describe *every* Chess960 board. It cases where the 'KQkq' castling flags cannot sufficiently encode the board, the column/file is used to record which rooks have castling rights.

Additionally, suppose a PGN has the following two headers:
  [Variant: Chess960]
  [FEN: 'rkr5/8/8/8/8/8/RKR5 KQkq w - 0 1']

Is this an invalid PGN that should be rejected (because the `Variant` has `Chess960`, but the castling flags are `KQkq`)?
If the PGN is accepted:
    1) Should the Variant header 'win' and the KQkq flags be converted to 'ACac'?
    2) Should the 'KQkq' flags in the FEN 'win' and the 'Chess960' header be ignored?
    3) Should the FEN be accepted with the understanding that the KQkq flags apply to the outermost rooks?

If option 3 is chosen, then suppose the following FEN is loaded via the `load()` method: 'r3k2r/8/8/8/8/8/R3K3R KQkq w - 0 1'. If 'KQkq' applies to the outermost rooks, is this a game that should be played with Standard rules or Chess960 rules?

Having said all that, I will be glad to implement whatever you decide. If you would prefer that `load()` not take a second parameter, then that will require that all X-FEN strings
use 'A-Ha-h' to encode the castling flags.



> Do you have any idea if the performance of the classical move generator is impacted by the 960 changes?

I will try some things and report back my findings.

> The sort on line 1617 ( const moves = this._moves().sort((a, b) => b.flags - a.flags)) has me a little concerned. 

Thanks to your question, I found a bug. The bug has been fixed which required removing the `sort()`.


Lastly, I apologize for interrupting the work that sowson has done. I don't mean to be stepping on his toes. I know handling PRs from one person is difficult enough. If handling the same feature from two people is too much, I won't take offense if you let me know.

deverac:
> Hi, @deverac. I love that you did the performance tests. Can I took to my PR #489 ?? :).

You are welcome to everything in the PR, and may copy and/or modify it in any way you desire.

deverac:
> And this one:
> 
> `it('X-FEN O-O 960 29', () => { const chess = new Chess( 'rnbkrbqn/p2p1p1p/2p1p1p1/1p6/8/BPN3PB/P1PPPPQP/R2KR2N w EAea - 0 11', { enable960: true} ) expect(chess.moves().length).toBe(36); chess.move({ from: 'd1', to: 'g1' }) expect(chess.fen()).toEqual( 'rnbkrbqn/p2p1p1p/2p1p1p1/1p6/8/BPN3PB/P1PPPPQP/R4RKN b ea - 1 11', ) })`

I believe that in the example above, you are trying to castle kingside with white. If that is the case, the move is incorrect because you are trying to castle the king by moving it 'over' the rook. To castle kingside with white, you must move the white king 'onto' the kingside rook, so the move should be `{from:'d1', to: 'e1'}`.

deverac:
> Hi, @deverac. Can you please check the test below?
> 
> `it('X-FEN O-O 960 28', () => { const chess = new Chess( 'rnbkrbqn/p1pp1ppp/4p3/1p6/8/BPN3P1/P1PPPP1P/R2KRBQN w AEae - 2 9', { enable960: true} ) chess.loadPgn('9. Bh3 c6 10. Qg2 g6') expect(chess.moves().length).toBe(36); chess.move({ from: 'd1', to: 'g1' }) expect(chess.fen()).toEqual( 'rnbkrbqn/p2p1p1p/2p1p1p1/1p6/8/BPN3PB/P1PPPPQP/R4RKN b ea - 1 11', ) })`

There are a few issues with the test above.

When PGN data is loaded, it resets the chessboard to standard starting position (so the FEN provided in the constructor is immediately overwritten). Because the chessboard is reset, the first move of the PGN `Bh3` fails because there is a white pawn on `g2` that blocks the move. This behavior existed prior to the Chess960 updates.

If the FEN in the constructor is manually updated with the moves specified in `loadPgn()`, it will result in the following FEN: `rnbkrbqn/p2p1p1p/2p1p1p1/1p6/8/BPN3PB/P1PPPPQP/R2KR2N w AEae - 2 9`. 

The `loadPgn()` line should be removed.

The move for castling should be `{from: 'd1', to: 'e1' }`, (rather than 'd2-g1').

The FEN in the `expect()` statement will have to be updated.
 1) The queenside castling flags, `ae`, are converted to `kq`. (This behavior may change, depending on what jhlywa decides.)
 2) The halfmove number must be changed to 3 from 2 (because the castling move was applied)
 3) The full move number must be reduced to 9 from 11 (because the PGN moves no longer exist)

If you do the above, the test should pass.

deverac:
Just FYI.

Castling in Chess960:
On [LiChess.org](https://lichess.org/variant/chess960) "The only way to castle is to move the King onto the Rook."
On [chess.com](https://support.chess.com/en/articles/8614060-how-do-i-castle-in-chess960) "click first on the King, then on the Rook."
On [reillyraab.com](https://reillyraab.com/fischer-random.html) "Castling is performed by dropping your king on top of a rook."

Castling flags (per [Wikipedia](https://en.wikipedia.org/wiki/Chess960)):
In X-FEN, the castling markings "KQkq" have their expected meanings: "Q" and "q" mean a-side castling is still legal (for White and Black respectively), and "K" and "k" mean h-side castling is still legal (for White and Black respectively). However, if there is more than one rook on the baseline on the same side of the king, and the rook that can castle is not the outermost rook on that side, then the file letter (uppercase for White) of the rook that can castle is used instead of "K", "k", "Q", or "q"; in X-FEN notation, castling potentials belong to the outermost rooks by default.

The solution implemented by chess engines like [Shredder](https://en.wikipedia.org/wiki/Shredder_(chess)) and [Fritz](https://en.wikipedia.org/wiki/Fritz_(chess)) is to use the letters of the columns on which the rooks began the game. This scheme is sometimes called Shredder-FEN. For the traditional setup, Shredder-FEN would use HAha instead of KQkq.

deverac:
Some thoughts about having `load()` accept only a single parameter. It seems that there are three options:

1. Require the castle flags to set the state. i.e. Any FEN string with castle flags `KQkq` will be interpreted as Standard position; any FEN string with castle flags `A-Ha-h` will be interpreted as Chess960 position.

2. Auto-detect whether the (X-)FEN string describes a Chess960 or Standard board position. There will be cases where the auto-detection will be wrong so there will have to be an optional second parameter to explicitly specify the type of game. The only benefit of this approach is that the second parameter would be optional most(?) of the time.

3. Add a prefix (or suffix) to the (X-)FEN which specifies the type of game. e.g. Anything that does not have the prefix will be interpreted as a Standard board layout. Anything with the prefix (e.g. `Chess960:`) will be interpreted as a Chess960 board layout. Loading a FEN string would look something like `load('Chess960: k7/8/8/8/8/8/8/8/K7 w - - 0 1')`

If you want `load()` to accept a single parameter, I think option 3 could work well.

If `load()` is modified to accept a single parameter, would you like the `enable960` option removed from the `Chess()` constructor? It seems the same sort of use-cases would apply to both situations.

(IMHO, some kind of header for the (X-)FEN string which identifies the type of game is an overdue requirement in the community of chess-variants.  Currently, the FEN string, and the type of game that the FEN string describes are two separate pieces of information that must be passed around together. A header would allow the remaining string to be fit-for-purpose instead of trying to shoe-horn every variant into the original FEN structure.)

deverac:
All of the suggestions seem sensible and are do-able. My (long-winded) thoughts are below.

* I would agree that some way of explicitly specifying the variant is preferable to trying to deduce the variant from the (X-)FEN.

* I'm not opposed to the `getCastlingRights()` suggestion. The function name makes perfect sense for standard chess (since booleans are returned), but is a slight misnomer if the squares are returned. I doubt the misnomer will be a problem for anyone using the `chess.js` library.

* If I was a user of a chess library that had a `reset()` method, I would assume that it would revert to some previous/initial position, not generate a completely new starting position (for 'chess960'). Also, having `reset()` take a parameter seems a bit strange to me. It suggests that I could construct a `Chess(FEN, {variant:'chess960'})` object, but then `reset({variant:`standard`})` it to a standard chess game. And vice-versa. That ability could be desirable in some circumstances, but seems like odd behavior to me.

* Allowing the user to specify the `variant` in the `load()` function could be tricky. PGNs may/should include the variant in their header. A decision will have to be made about how to proceed when the variant supplied by the PGN conflicts with the variant supplied by the `load()` function.

* `getVariant()` vs. `isVariantChess960()` are two different ways to accomplish the same thing. e.g.:
```
if (lib.getVariant() == 'chess960') {
   // do something
}
```
vs.
```
if (lib.isVariantChess960()) {
   // do something
}
```

The first case (`getVariant()`) means the 'chess960' string has to appear throughout the code and seems more error-prone due to typos. (e.g. `lib.getVariant() == 'Chess960'` will never be true). Also, if `getVariant()` is used, I would bet that the first thing every programmer will do when using `chess.js` is create the helper method: `function isVariantChess960() { return lib.getVariant() == 'chess960' }`

The second case (`isVariantChess960()`) allows auto-complete and introspection, which can be helpful when coding. But it also means that the `chess.js` library must have a separate function for every variant that is supported, which could clutter the API.

* There may be a need for another method which returns all supported variants. Alternatively, the variants could simply be noted in the documentation and it would be up to the programmer to check the documentation to determine what variants are supported.

Someone (not me) will have to decide what the API for `chess.js` should be. Currently, `chess.js` supports a 'standard' chess game. This PR would add the 'chess960' variant. There are many chess variants (different rules, different pieces, different board sizes). As soon as one variant is introduced, surely requests to support other variants will follow.

It may be decided that `chess.js` will only support 'standard' chess and that anyone wanting more will have to find a different library. For anyone wanting to play chess variants, the [Fairy Chess](https://fairy-stockfish.github.io/)  project may be what they are looking for.

I'm OK with refactoring this PR to suit the API goals of `chess.js`. I'm also OK with having this PR rejected if it is 'out-of-scope' for the goals of `chess.js`. So, if any changes to this PR are desired, please let me know and I'll do my best to accommodate them.

neofight78:
Thanks for your detailed reply!

@jhlywa will have the final say on what he thinks the API should be, so I would hold off on any changes until he next gets a chance to look at this. I'm certainly open minded on most of the possible choices.

I do know that he does want to add Chess960 support (there are lots of requests for it). We just want to make sure we get the API right as that's more disruptive to change later on after people have started to depend on it.

You make some good points. My main concern is that we don't add lots of variant specific methods as far as is possible.

You are right that the getCastlingRights() suggestion is a little funky, if we were starting from scratch maybe there would be something more elegant that could cater for both variants. Perhaps in v2... I still like the suggestion though.

I think you are probably right about reset() it should just go back to the starting position of the given game. And not change the type, as none of the headers etc get reset. However I think in your PR  reset() is not changed, but if it's a 960 game it should probably go back to that specific 960 start position or at least a random position? Going back to the standard start position would be weird, I think.

I'm not proposing specifying the variant when loading a PGN, as you rightly point out that could result in confusion. Rather when loading a FEN as we need to resolve the ambiguity there. Maybe that too is not necessary given that the variant will already be set at that point? (in contrast to the constructor).

I'm proposing a Variant type, so typos aren't possible without getting a type error, you also get IDE autocomplete with that too. It just seems cleaner / more future proof to me. If we never ever add a new variant other than 960 then maybe a boolean / 960 specific method is ok.

SymboLinker:
Thanks for this pull request! I'm already using this :) 

I really like the simple API of this repository and I wanted chess960. This branch contains a working 'chess.ts' file for this, hooray!

Perhaps accepting the proposal with convenient methods like `.isVariantChess960()` is fine for now? If it changes, everyone can check the changes in the [Releases tab](https://github.com/jhlywa/chess.js/releases/) before updating.

neofight78:
Great to seem someone is using this branch already!

@deverac I checked with @jhlywa and I don't think we are going to support anything other than standard and 960, so you can disregard my comments about variants etc. Here are my latest thoughts:

* Rather than enable960 and variantChess960 - it would be a bit more consistent and a little shorter to just use chess960 everywhere

* Chess([ fen ], { chess960 }) - tweak naming  

* getCastlingSquares() - actually I've come round to this idea, but it would be better to return undefined rather than an empty string for when there is no castling right.

* isVariantChess960 () -> isChess960()

* setVariantChess960() -> setChess960() - I also wonder if this is needed / makes sense? Seems like something that should only be set when starting / loading a new game?

* generateChess960Fen() - we need to document whether this will generate X-FEN or Shredder-FEN.

* validateFen(fen, {  chess960 })  - tweak naming  

* reset() - needs to go to relevant starting position (standard or the specific 960 starting position the game started with)  - is possible we have a chess960 pgn with no recorded starting position though?

* clear() - seems to reset everything so should probably accept a { chess960 } parameter

* fen() -  we need to document whether this will generate X-FEN or Shredder-FEN for chess960 games. possibly accept a parameter for which is desired?

Keen to hear your thoughts? Also @jhlywa what do you think?

@deverac if you have time it would be great if you could rebase this PR so that it's in a merge-able state again.

deverac:
@jhlywa @neofight78 Thank you for the guidance. I'll get to work on incorporating the suggested changes and rebasing. I was glad to see a definitive answer about whether other variants might need to be supported. (Answer: No)

deverac:
@neofight78 Thanks for the catch. Are there other commands I should run prior to pushing a change?

I've still got some work to do on this branch to incorporate the suggested code changes. Since this PR was months old, I wanted to get it 'rebased' first.

neofight78:
@deverac Let us know when you're done and we'll have a look through it.

deverac:
@jhlywa, @neofight78 I've updated the PR and it's ready for review, at your convenience.

I believe that the 'Chess 960 PGN import castling error' issue noted by GuillaumeSD is due to an unsupported/invalid PGN Header, not due to a bug with `chess.js`.

deverac:
@neofight78 
I like your idea of a single `getCastlingRights()` function (rather than having two methods that do almost the same thing). If @jhlywa also agrees, I'll update the code and documentation.

jhlywa:
Great stuff @deverac!  I agree lets go with `getCastlingRights()` vs two similarly named/purposed methods.

deverac:
@neofight78, @jhlywa 

An updated PR is ready.

 * Merged `getCastlingSquares()` and `getCastlingRights()`. (The former name  was removed.) This was trickier than anticipated because  previously `getCastlingRights()` was only concerned with rights. The updated  function had to handle both squares and rights. 

 * `getCastlingRights()` now returns squares or `undefined` values, so updated tests to expect `Truthy()` and `Falsy()`, rather than `true` and `false`.

 * Removed some code where I was trying to be too clever by restricting castling-rights based on if both sides had kings and rooks directly across from each other or not. This restriction is pointless (and wrong) when castling-rights can be set directly.

FYI. There is a 'quirk' (which I documented in `index.md`) related to setting the castling-right on an inner-rook. A Chess960 game may have (e.g.) two or more same-color rooks on one side of the king. Only one of them is allowed to have a castling-right. The current `chess.js` API does not allow specifying which rook `setCastlingRight()` should operate on (currently, the outer-rook is always operated on). In the rare case of a pre-existing Chess960 game where there are two or more rooks on one side of the king and the user wants to set the castling-right on an inner-rook, the user must remove the outer-rook(s), set the castling-right (on the previously inner-rook) and then re-add the outer-rook(s).

If this 'quirk' is unacceptable behavior, let me know how such cases should be handled and I'll update the code.

In cases where the user is manually setting up a Chess960 game from scratch and an inner-rook needs a castling-right, the inner-rook can be placed on the board first, then the castling-right can be set, then any other rook(s) can be placed to the left or right of the king in any order.

neofight78:
It would be good to get some kind of symmetry with getCastlingRights() if possible.. Perhaps make it so that you can pass either a boolean or a column when disambiguation is needed. Maybe getCastlingRights() should return columns rather than squares too?

neofight78:
One thing that occurs to me is that the starting position is symmetrical, so that any disambiguation applies to both sides. So I think just file makes sense for both methods. If b1 is the castling rook square for example, then b8 is necessarily the castling rook square for black.

deverac:
Thank you for those pointers about the documentation. `npm check` was passing on my machine, but, as noted, it failed on the build machine. Re-pulling and reinstalling the node modules solved my issue and I was able to see the same errors that the build-machine reported. I've updated the PR and the build should (fingers crossed) now pass.

I really like the `_rooks` suggestion that you made. It's much nicer than having to constantly loop through a two-element array and it cleaned up the code quite a bit.

It's also great to now be able to see the changes to the public API.

I've still got a bit of work to do and will post an update when progress is made.

deverac:
Generalizing the premove/preprocess stuff was a great suggestion! Thanks. The code is much less opaque now (at least to my eyes). I took a crack at implementing your suggestion, but if it's not quite what you wanted, please let me know.

The `_rooks` field is now deep-copied, rather than being serialized.

If you see anything else that needs tweaking, updating, renaming, etc., please mention it, because now is the best time to get it done.

With all the changes and the merge conflict, Rebasing was going to be painful. I created a new 'clean' PR of all the changes that were made in this PR. The new PR is #551.

Unless you object, I would request that this PR be closed and that any further updates be done on the new PR (#551).

neofight78:
Great work, thank you! That's the main two things that jumped out at me covered. Probably we will want to merge at some point soon to avoid further conflicts etc I'll have another look at the code and continue the conversation on the new PR...

PR #497, opened 2025-03-25
tests: increase coverage

axolotl-logic:
I added a target for generating coverage reports and then worked in a couple extra cases into the tests.

FYI, each new test should have added additional coverage.

PS. You can find the report here after running `npm run test:coverage`: `coverage/lcov-report/index.html`. It was already excluded by gitignore.

### DIFF
--- a/__tests__/undo.test.ts
+++ b/__tests__/undo.test.ts
@@ -11,9 +11,3 @@ test('undo - works', () => {
 
   chess.undo()
 })
-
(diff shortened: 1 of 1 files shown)


axolotl-logic:
I also worked in a case that's mentioned in comments but not actually tested against.

jhlywa:
Great stuff @axolotl-logic! Thanks for the PR!

PR #500, opened 2025-03-29
test auto-updating of Setup/FEN tags

axolotl-logic:
I had this in another PR, but accidentally deleted it.


jhlywa:
Thanks again @axolotl-logic!

PR #509, opened 2025-05-03
Parse PGNs with peggy

neofight78:
Uses peggy to parse pgn files. It already returns a full game tree, so if and when this PR is merged, I can look at adding RAV support. All existing tests pass.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+src/pgn.d.ts
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,3 @@
 coverage/
 package-lock.json
+src/pgn.*
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "eslint": "^8.12.0",
         "jest": "^27.0.6",
         "jest-extended": "^2.0.0",
+        "peggy": "^4.2.0",
         "prettier": "^3.1.0",
         "ts-jest": "^27.0.4",
         "typescript": "^4.6.3"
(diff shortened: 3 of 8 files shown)


jhlywa:
Thanks for the much needed PR @neofight78!

Two quick questions:

- It seems like the [double annotation tests](https://github.com/jhlywa/chess.js/blob/9290c640923c088645b38975fc943ccf7c5ce345/__tests__/load-pgn.test.ts#L168) should be failing given the current grammar for `move`. Should we change the `suffixAnnotation?` to `suffixAnnotation*`?  Some of the code-paths run `strippedSan(move: string)` which may be masking some intended behavior, or more likely, just confusing me :)

- I'm curious about any performance hits we may experience when using peggy for parsing.  Do we have a large PGN dataset with could parse with both `v1.2.0` and the peggy branch? I'm really crossing my fingers that the impact is minimal, because this grammar-based approach really simplifies the code and make future RAV support pretty straightforward.

Thanks again and I can't wait to merge this!

neofight78:
I'm not sure why you think they should fail? Multiple suffix annotations are not possible. Although a suffix annotation may be 1 or 2 characters as covered by the grammar. If you are thinking of check and checkmate symbols they are captured as part of the move itself, as there aren't really annotations as such. Or maybe I am missing something?

The test suite runs in about the same time, but I've not done any performance testing beyond that. I'll grab a large PGN from somewhere and report back...

neofight78:
Ok, so I hacked together a script to parse around 5500 games with no annotations. Roughly speaking the new parser was about 3% slower. I tried a file of 10 annotated games and parsed those 1000 times, with similar results. Given the new parse actually does a bit more work in returning a full game tree, that's not bad I feel.

neofight78:
Fixed the build so that it copies the parser code over to the dist folder. Plus a small tweak to clean and gitignore.

jhlywa:
Awesome work @neofight78!  Can you please merge this PR as well as the PGN benchmarks?

PR #512, opened 2025-05-07
Allow user to always include enpassant square in fen()

sblackstone:
Adds an option to fen() which will include the enpassant square regardless of whether the option can actually be exercised or not on the next move.

My use case for this is compatibility with another library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,10 +228,12 @@ chess.fen()
 // -> '8/8/8/8/8/8/8/8 w - - 0 1' <- empty board
 ```
 
-### .fen()
+### .fen({ forceEnpassantSquare = false) = {})
 
 Returns the FEN string for the current position. Note, the en passant square is
 only included if the side-to-move can legally capture en passant.
+
+The enpassant square will always be included if forceEnpassantSquare is true.
 
 ```ts
 const chess = new Chess()
--- a/__tests__/fen.test.ts
+++ b/__tests__/fen.test.ts
@@ -40,3 +40,14 @@ test('fen - ep square only if en passant is legal (illegal - pinned - #2)', () =
     'rnb1kbn1/p1p2p2/PpPp2qr/4pPp1/8/R1P4p/1PK1P1PP/1NBQ1BNR w - - 0 2',
   )
 })
+
+test('fen - allow EP square to be included by option', () => {
+  // black queen pins the ep pawn, making ep illegal (submitted by @ajax333221)
+  const chess = new Chess(
+    'rnb1kbn1/p1p1pp2/PpPp2qr/5Pp1/8/R1P4p/1PK1P1PP/1NBQ1BNR b - - 0 1',
+  )
+  chess.move('e5')
+  expect(chess.fen({ forceEnpassantSquare: true })).toEqual(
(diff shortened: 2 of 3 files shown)


sblackstone:
I actually wrote it that way intentionally for readability, versus....

```
if (epSquare !== EMPTY) {
  if (force) {
     // force
  } else {
    // gigantic indented block... 
  }
}
```
I'm happy to do this instead if you prefer...

neofight78:
You end up with a gigantic indented block either way (by virtue of it being in an if statement). IMHO it's a little easier to read the intent with the way I'm proposing.

jhlywa:
I think this looks fine, but how should we handle other functions that emit FEN (e.g. `moves({verbose: true})` or `history({verbose: true})`)?

We could obviously add an additional parameter to each function ... but there's an implicit dependency between these parameters (e.g. you would never just call `moves({forceEnPassantSquare: true})`). There's also the possibility of including some sort of object-wide configuration to handle these cases ... `.setConfiguration({forceEnPassantSquare: true, ... })` or something similar.

I'm honestly not sure which approach is better.  Keeping the API clean and ergonomic has always been a challenge.

sblackstone:
> You end up with a gigantic indented block either way (by virtue of it being in an if statement). IMHO it's a little easier to read the intent with the way I'm proposing.

It's fine, I'll switch it.  My personal preference is to avoid too many levels of indention, it already has an if -> loop -> if -> if so it ends with 4 braces... What's one more :-)

neofight78:
@jhlywa Good point about needing to update those other functions.

My instinct is that the format of the FEN is not intrinsic to the game, it's a requirement of the client to get output in a certain format, so should be an option that is passed in as needed.

Contrast that with the Chess960 PR (where we also have discussions about the API). There I feel it is a property of the game and therefore should be one way or another be configured on the game level.

Whether we ever need to configure something on the level of the Chess object itself I am not sure so.

Those are my current thoughts at least...

sblackstone:
There's no reason we can't do both, have a global config and a function option

sblackstone:
There's no reason we can't do both, have a global config and a function option

jhlywa:
I agree @neofight78. I don't feel like this setting is intrinsic to the game, and a referentially transparent API is always nicer IMO.

I scanned the code and It looks like `getComments()` and `removeComments()` also return FEN strings. That brings the current count to five methods that would need this optional `forceEnPassantSquare` parameter. I'm a little worried about this thing rippling through the API, especially if any future methods produce FEN. 

Or ...... we can just punt and only implement it for `fen()` proper.

neofight78:
@jhlywa I don't know if that would be so bad? But certainly if just having it for fen() fixes @sblackstone's case then that seems a practical decision. The parameter could always be added elsewhere if it turns out it's needed. Although TBH I expect this will not be used much?

The other option would be to push back and say it's for the other library to fix their code. Or alternatively, and it would be a little fiddly to do, but an adapter could look at the last move and update the ep square in the fen as a workaround.

jhlywa:
Yeah @neofight78, let's just take the practical approach and implement this for `fen()`.  I think the PR is ready to merge as is.

sblackstone:
Should be up to date now.

PR #516, opened 2025-05-13
Added support for chess annotation symbols

brobert1:


neofight78:
Great, thanks for that. A few bits of feedback:

* I'm not sure why you've added the extra symbols. The only 6 suffix annotations allowed by the PGN spec, and recognised by the parser are: !", "?", "!!", "!?", "?!", and "??"

* There is one failing test because there is now  a suffix property that it wasn't expecting before. So that test just needs updating.

* The rawSuffix/suffix logic is not needed, there are can only be one suffix, again from the list mentioned above.

* getComments() won't return a suffix if there is no comment for that position / move.

brobert1:
Hi there, @neofight78!
I have implemented the requested feedback. Please let me know if there is something else you would like to change.

neofight78:
Thanks @brobert1 !

Two problems I spotted:

1) getComments() won't return a suffix if there is no comment for that position / move. (same as before)
2) !!, !? etc are getting split into two characters, this is a bug in the parser grammar. update the suffix allocation rule in the peggy file to fix this, by adding $ at the start of the rule, i.e. (don't forget to run "npm run parser" to rebuild afterwards).

suffixAnnotation "suffix annotation"
	= $[!?] |1..2|

brobert1:
Hello again, @neofight78!
I have updated the parser grammar and solved the issue + getComments(). However, I am not completely sure if this is the best approach regarding the method. Please let me know your opinion and, if possible, a suggestion so I can better understand your vision regarding how getComments() should behave.

neofight78:
Thanks, this seems to work quite well now.

Let me know what are you misgivings about the approach?

Maybe if there is no comment it's better not to return one rather than a blank string? Or perhaps it would be better to have a separate getSuffixes() method and not mix the two together?

@jhlywa What do you think of the API?

brobert1:
To be fair, this is the best approach I’ve found to make it work correctly. I only had second thoughts because I wasn’t sure I fully understood what you asked me to do. As far as I’m concerned, this PR addresses the initial issue and can be merged, unless you have any feedback or modifications to suggest.

jhlywa:
Sorry for the delay. A couple of thoughts:

- Is the term `suffix` (as used in the API and return values) specific enough to indicate its intent? To me `annotation` might be a more descriptive term, but maybe it's just me.
- Could we add the new methods to the README docs?

brobert1:
Ok so what is left to do so you can merge this PR? Update the README?

jhlywa:
I think `suffixAnnotation` is perfect, let's refer to it as that in the API and `getComments()` return value.

Since this will be a breaking change, we'll need to postpone its inclusion until the next major release. We may be able to merge it into master sooner if we agree that the upcoming 1.3.0 release will be the last v1 release - after which you could pull from Github until we publish v2 to NPM.

brobert1:
Hello, @neofight78! Yes, i will try to implement this changes asap so they are ready for merge.

brobert1:
Hello, @neofight78! I am solving the issues right now so you can merge it right away when ready.

neofight78:
Thanks @brobert1! I think the code is looking good now! There is just one (new) final step to do.

As part of the build process we now check if the public api has changed. This is so we can make sure that we don't accidentality change or break things. This is gets run as part of `npm run check`, but you can run it separately with `npm run api:check` too.

If you want to view any of the differences that are picked up, do a diff of the reports in temp and etc. To confirm you are happy with the changes use `npm run api:update`. If you check it again you should see there are no reported changes. Commit and push the updated report and we should be good to go!

We''ll have the process documented soon (see PR #548)

neofight78:
Thanks @brobert1!

@jhlywa I think this one is good to go now.

brobert1:
Hello, @neofight78  @jhlywa!
Do you have any idea when the changes regarding suffixAnnotations will go live? I’m asking because we are looking forward to implementing it in our app, and these updates are quite essential for our use case.

Thanks in advance!

neofight78:
To quote @jhlywa's previous answer to this question:

> Since this will be a breaking change, we'll need to postpone its inclusion until the next major release. We may be able to merge it into master sooner if we agree that the upcoming 1.3.0 release will be the last v1 release - after which you could pull from Github until we publish v2 to NPM.

So I will merge to master now, and you can install from git until we make the next release. I think that is probably several weeks away at least and it will be the first alpha version of v2.

PR #517, opened 2025-05-18
Zobrist hash

neofight78:
Adds a new method call hash() which will return a 64bit bigint.

* Benchmark performance about the same, maybe very slightly faster.
* The internal ep square is now only set if there a pawn able to capture next move (does not check full legality, checks etc)
* The threefold repetition tracking is now done with a map of hashes 
* Hash integrity is mainly checked during the move and load pgn tests
* Standard zobrist implementation, with included stable/seeded rng()

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,11 +340,11 @@ chess.getHeaders()
 
 ### .hash()
 
-Returns a unique 64-bit bigint hash for the current position.
+Returns a unique 64-bit hash as a hexidecimal string for the current position.
 
 ```ts
 chess.hash()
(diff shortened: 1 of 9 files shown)


jhlywa:
Great stuff @neofight78.

One minor thing, I don't think I've ever personally worked with BigInt while using JS, and I'm guessing others haven't either. How would you feel about changing the external `.hash()` API to produce a hex-encode string, ala git hashes? I think using BigInt internally is fine.

jhlywa:
Looks good!  Merged

Issue #113, opened 2016-04-27
Cannot parse a pgn game not starting in initial position.


ppipper:
I am not sure if its a bug. This game cannot be parsed unless you manually change move numbers to 1, 2, 3 etc.

Is this a bug?

[Event "SampleSite"]
[Site "?"]
[Date "2008-01-04"]
[Round "2"]
[White "Zhou Jianchao"]
[Black "Fidaeyee, H."]
[Result "1-0"]
[BlackElo "2074"]
[FEN "1n1r3r/1k2q1p1/1bp1p2p/5p2/2Pp1B2/5QN1/5PPP/RR4K1 w - - 0 29"]
[PlyCount "1"]
[SetUp "1"]
[SourceDate "2012.04.30"]
[WhiteElo "2551"]

29. Rxb6+ Kxb6 30. Qb3+ Qb4 31. Qxb4# 1-0

ryanhs:
hmm.. good case,

but as far as i know, in SetUp PGN, its always from 1.
never find from 29, because software doesn't know movements from 1 - 28

jhlywa:
I'll have to consult the PGN spec to see how to handle this.

ppipper:
Thanks.
By the way, the pgn text above can be pasted into some commercial programs (aquarium, chessbase) and it seems it is properly parsed, with the starting move on 29. Rxb6 and so on.

jhlywa:
This was fixed in #150.  I'll leave this issue open until I add test for it.

sanichi:
I think the problem is that the regex that parses out the `header_string` in `load_pgn()`, assumes the game begins with move number 1. Specifically it is trying to match `1.` (the digit 1 followed by any character) on line 1433. I guess, however, what was intended was: `1\.` (the digit 1 followed by a full stop).

But, since game fragments (PGN with a FEN) don't always begin with move 1, or a move number which starts with 1, what should be here is `[1-9]\d*\.` (a positive integer followed by a full stop).

Anyway, I've tried this and it seems to work.

axolotl-logic:
> This was fixed in [#150](https://github.com/jhlywa/chess.js/pull/150). I'll leave this issue open until I add test for it.

Not sure if you're still waiting on writing that test, but here's a start (I left expect empty).

```typescript

  it('Github Issue #113 Cannot parse pgn not starting in initial position', () => {
    const chess = new Chess()
    const pgn = [
      '[Event "SampleSite"]',
      '[Site "?"]',
[...]

PR #523, opened 2025-05-25
Added regression test for parsing of move numbers

neofight78:
Regression test for #113 

@axolotl-logic This test checks the user's pgn loads correctly.


axolotl-logic:
Thank you so much!

PR #525, opened 2025-05-30
Fix castling with check

neofight78:


neofight78:
Yes, if it's not too much trouble.

jhlywa:
Tagged and pushed

PR #527, opened 2025-06-05
Use rollup to build dist files

### LINKED ISSUE #526: pgn.js dependency vs. self-contained chess.js
i am using chess.js ESM module by importing from a local copy of e.g. https://cdn.jsdelivr.net/npm/chess.js@1.2.0/dist/esm/chess.js
starting from version 1.3.0 chess.js depends on pgn.js, meaning i need to maintain two local files as well as deal with fixing the pgn import.
would be convenient to have the option of a single-file self-contained chess.js
btw, thanks for this great lib!

neofight78:
Uses rollup to bundle the peggy generated js and node.ts files with the main chess.ts file, making it a one file library again. This should fix #526

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,17 +9,21 @@
       "version": "1.3.1",
       "license": "BSD-2-Clause",
       "devDependencies": {
+        "@rollup/plugin-commonjs": "^28.0.3",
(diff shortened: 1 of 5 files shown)


neofight78:
Looks like jest is having problems...

PR #528, opened 2025-06-06
Use rollup for build and vitest for tests

### LINKED ISSUE #526: pgn.js dependency vs. self-contained chess.js

neofight78:
Uses Rollup so we can bundle code to get back to a single js file again. Fixes #526 

Rollup means we need to build the peggy parser as esm and that means we need vitest to support running tests with esm (couldn't get it working with jest). vitest is a nice upgrade anyway and speeds the tests up a little.

Rollup doesn't work with node 14 so took that out of the CI build, as it's no longer maintained anyway. While I was at it, I added node 24 in.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -15,7 +15,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 20.x, 22.x, 24.x]
+        node-version: [18.x, 20.x, 22.x, 24.x]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
 
     steps:


neofight78:
Turns out vitest won't run on node 16, so removed that as well.

neofight78:
Having said all of that, node 14 & 16 can't be used for building the library, but it shouldn't affect which versions can consume the library. Just that, as we are not running the tests, we can't guarantee it.

jhlywa:
Good stuff @neofight78! I'm fine saying you need node 18+ for building the library. No sense in bending over backwards to support outdated versions.

PR #539, opened 2025-06-22
isPromotion method for UIs

neofight78:
Inspired by #148, covers off issues raised about knowing when a UI needs to prompt for a promotion piece.


jhlywa:
Thank you @neofight78!  Sorry for the delay getting these merged

PR #540, opened 2025-06-22
Improve visibility on npmjs.com

neofight78:
In the search on npm the chess.js entry doesn't look great as it shows the markup for the various badges:

image

There are also no keywords setup.


neofight78:
Also added benchmarks and  website folders to .npmignore.

PR #541, opened 2025-06-23
Add update version script

neofight78:
Follow up to #532. You can now run something like this

```npm run update-version 1.4.1```

and it will update the package version and create and configure a snapshot of the docs.


neofight78:
NB Adding node types required a newer version of TS, which then needed a newer version of eslint etc...  Hence additional commits.

PR #543, opened 2025-06-28
Add api-extractor to track api changes

neofight78:
In the upcoming changes for v2, there will be deliberate changes to the public api, but also accidental changes could creep in as part of any code reorg, so I propose using api-extractor to help ensure there are no unintentional breakages, and all necessary types are exported etc.

```npm run api:check```

Will check to see if there have been any changes to the public api. If there any any unintentional changes or breakages they can be fixed and verified. If changes are all anticipated / ok, then running:

``` npm run api:update```

Will update the gold api file so that the changes will be accepted.


neofight78:
@jhlywa rebased with latest changes + small config fix

PR #550, opened 2025-07-20
Break circular dependency between Move and Chess

neofight78:
This is technically a breaking change although I suspect nobody is using the Move constructor as InternalMove is not public. A prerequisite to upcoming RAV changes.


neofight78:
Also added missing isNullMove() method

PR #553, opened 2025-08-20
Improve Draw Rule Handling

ANDREA-1720:
> [!IMPORTANT]  
> The changes applied by this PR **can** be described as **BREAKING CHANGES** as contains changes to the  _public API_ as descibed in the **_Changes_** sections

### Overview  
This PR refines the implementation of draw conditions to make them more compliant with official FIDE chess rules.

### Changes  
- **Manual Drawing**
  - Added `draw()` method to allow players to manually claim a draw.

- **Fifty-Move Rule**
   - Split logic into:
    - `canDrawByFiftyMoves()` → detects when the position qualifies for a draw claim.
    - `isDrawByFiftyMoves()` → requires explicit claim (`_isManuallyDrawn`).
  
- **Threefold Repetition**
  - Split logic into:
    - `canThreefoldRepetition()` → detects when the position qualifies for a claim.
    - `isThreefoldRepetition()` → requires explicit claim (`_isManuallyDrawn`).
  - Updated tests (`can-threefold-repetition.test.ts`) to reflect new semantics.
  
- **Null Move Handling**
  - At an high-level it does not change, only the implementation changes.
  - Added `_nullMovesCount` to track consecutive null moves ( with `"--"` SAN).
  - Game considered drawn if `canThreefoldRepetition()` **and** 6 consecutive null moves occur.

> [!NOTE]  
> The handling of **manual draws** must be implemented by the developer using the library, for example:  
> - For draws not triggered by fifty-move or threefold repetition rules, the developer should ensure that **both players agree** to a draw.  
> - For draws due to fifty-move or threefold repetition, it is sufficient that **only one player requests the draw** for it to be valid.

### Why  
- Aligns draw logic with **FIDE chess rules** (distinction between automatic and claimable draws).  

### Tests  
- Updated and renamed tests for `canThreefoldRepetition()`.  
- Confirmed compliance with all tests.

### DIFF
--- a/__tests__/can-threefold-repetition.test.ts
+++ b/__tests__/can-threefold-repetition.test.ts
@@ -2,29 +2,29 @@ import { Chess } from '../src/chess'
 import { split } from './utils'
 import { expect, test } from 'vitest'
 
-test('isThreefoldRepetition', () => {
+test('canThreefoldRepetition', () => {
   /* Fischer - Petrosian, Buenos Aires, 1971 */
   const fen = '8/pp3p1k/2p2q1p/3r1P2/5R2/7P/P1P1QP2/7K b - - 2 30'
   const moves = split('Qe5 Qh5 Qf6 Qe2 Re5 Qd3 Rd5 Qe2')
 
   const chess = new Chess(fen)
   moves.forEach((move) => {
-    expect(chess.isThreefoldRepetition()).toBe(false)
+    expect(chess.canThreefoldRepetition()).toBe(false)
     chess.move(move)
   })
-  expect(chess.isThreefoldRepetition()).toBe(true)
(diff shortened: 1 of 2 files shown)


neofight78:
I'm not necessarily ruling out changes, but I do think some confusion arises from the fact the a the Chess object both represents a game and the current board state. I think perhaps we need documentation here.

The Result header tag is the correct place for storing and retrieving the actual result of the game.

The methods such as isDraw(), isThreefoldRepetition() etc are implemented as you would typically find in a chess engine where draws get automatically get claimed.  So they are more about enforcing rules and evaluating the position etc, and not so much about giving the result of a human vs human game.

meruiden:
this would be awesome, currently very inconsistent with modern chess apps like chess.com and lichess.

PR #559, opened 2025-09-13
Fix: handle invalid FEN when side to move is already delivering check

### LINKED ISSUE #558: Missing invalid case in validateFen
It seems that `validateFen()` currently treats some invalid FEN strings as valid.
Specifically, when the side to move is already in check, the function still considers the FEN valid.

For example, the following FEN should be considered invalid, but is currently treated as valid:
`1n6/8/2Q5/1rK5/8/8/8/5k2 b - - 0 1`

Image

```javascript
import { validateFen } from "chess.js";

const fen = "1n6/8/2Q5/1rK5/8/8/8/5k2 b - - 0 1";

const isValid = validateFen(fen); // { ok: true }
```

YusungAN:
Fixes #558 

This PR updates the validateFen() function to correctly handle cases where the side to move is already delivering check.

### Changes
* Updated `validateFen()` logic to detect when the side to move is already delivering check
```javascript
export function validateFen(fen: string): { ok: boolean; error?: string } {
  
  // .. Rest of code

  // 12th criterion: is side to move already delivering check?
  const tokensTemp = tokens.slice()
  tokensTemp[1] = swapColor(tokens[1] as Color)
  const tempFen = tokensTemp.join(' ') // swap the side to move

  const tempChess = new Chess(tempFen, { skipValidation: true })
  if (tempChess.inCheck()) {
    return {
      ok: false,
      error: 'Invalid FEN: side to move is already delivering check',
    }
  }

  return { ok: true }
}
```
* Added and updated related test cases
  * Two testcases were added in `validate_fen.test.ts`.
  * Four testcases in `put.test.ts` were updated because they triggered the invalid FEN error. Please check if the modified testcases still reflect the original intention.
`put - occupying white en passant square clears it`
`put - replacing black pawn clears white en passant square 2`
`put - occupying black en passant square clears it`
`put - replacing white pawn clears black en passant square 2`

### DIFF
--- a/__tests__/put.test.ts
+++ b/__tests__/put.test.ts
@@ -154,7 +154,7 @@ test('put - occupying white en passant square clears it', () => {
     'rnbqkbnr/pppppp1p/8/8/3PPPp1/8/PPP3PP/RNBQKBNR b KQkq f3 0 3',
   )
 
-  chess.put({ type: KNIGHT, color: BLACK }, 'f3')
+  chess.put({ type: BISHOP, color: BLACK }, 'f3')
   expect(chess.moves()).not.toContain('gxf3')
 
   expect(chess.hash()).toEqual(new Chess(chess.fen()).hash())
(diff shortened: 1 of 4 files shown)


neofight78:
The code is creating a temporary chess object to do its validation, which, if we are calling it from the constructor is unnecessarily inefficient. 

We could add this validation to the constructor instead, but then we'd lose the functionality from this function.

So whilst I understand the aim, I'm not quite sure of the best way forward yet. @jhlywa Thoughts?

pabrams:
If there are some use cases where you care whether the FEN represents a legal position, and other use cases where you just want to check if the notation adheres to spec, maybe just have two different methods and people can use the one they want. Like `validateFen` and `validateFenPosition` or something. The latter calls the former, and does some extra processing to see if the position is actually legal.

PR #563, opened 2025-09-21
Move 'addMove' stateful function + add extra conditional to promotion check

joachimbulow:
## Moved stateful function into chess class
Having state manipulating functionality exported and separate from the chess class seems like it breaks the general structure of having state-less util-like functions exported outside of the `chess` class.

The function is not part of the public api - so should be a minor change

### DIFF
--- a/src/chess.ts
+++ b/src/chess.ts
@@ -650,42 +650,6 @@ function getDisambiguator(move: InternalMove, moves: InternalMove[]): string {
   return ''
 }
 
-function addMove(
-  moves: InternalMove[],
-  color: Color,
-  from: number,
-  to: number,
-  piece: PieceSymbol,
-  captured: PieceSymbol | undefined = undefined,
-  flags: number = BITS.NORMAL,
-) {
-  const r = rank(to)
-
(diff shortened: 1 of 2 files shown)


src/chess.ts
```diff
@@ -1680,6 +1644,46 @@ export class Chess {
...
+    color: Color,
+    from: number,
+    to: number,
+    piece: PieceSymbol,
+    captured: PieceSymbol | undefined = undefined,
+    flags: number = BITS.NORMAL,
+  ) {
+    const r = rank(to)
+  
+    if (piece === PAWN && ((color === BLACK && r === RANK_1) || (color === WHITE && r === RANK_8))) {
```
joachimbulow: We now check `color` + `rank` as opposed to just checking ranks

neofight78:
The changes / tests are unnecessary, as a white pawn can never move to the first rank, and a black pawn can never to move to the eighth rank.

joachimbulow:
Right - I am looking to extend the system to allow for more rules - which is what led me to that change.
For a static chess engine like this one, the infererred logic is just fine.

I have edited the description and removed the redundant check.

Issue #566, opened 2025-10-04
Split up the monolithic chess.ts file


JeffML:
I don't know if there is a reason to have everything in one file; if there is, I'd like to  hear it. 

I was able to put the Move class in a Move.ts file.  To avoid circular dependencies between chess.ts and Move.ts, some constants used by both are now in a types.ts file.  All tests pass, and builds with no issues.  Backward compatibility is preserved.

There were some minor changes to the API, namely **BITS** is exported, as well as **FLAGS** and **InternalMove** .  

I'll submit a pull request.

JeffML:
#568  supercedes #567

PR #567, opened 2025-10-04
refactoring Move from Chess.js to its own file

JeffML:
Created types.js to avoid circular dependencies; some changes noted in chess.js.api.md

Minor API changes.

#566


JeffML:
This PR is superceded by #568

Issue #561, opened 2025-09-15
Support multi-game PGN


JeffML:
There are a couple of things to consider. First is backward compatibility with `loadPgn()`; second is efficiently handling large PGN files with hundreds of thousands of games. 

#### Some ideas:
loadPgn() takes additional start/count parameters (defaulting to 0 and 1) which are the index and length of games to parse from the PGN file.

-  'light' parsing of the PGN file would separate the games without parsing each one fully
-  in the default case (start=0, number=1), loadPgn() behaves as it does now, preserving backward compatibility
-   in all other cases, loadPgn() returns an array of parsed games beginning at `start` and continuing to `number` of games
  
If `start` is greater than # of games, an error is thrown. If `number` is greater than # number of games, return array is truncated

Thoughts?

JeffML:
I glossed over some things.

There would be a Game class that handles just about all the functions that the Chess class does currently. To maintain backward compatibility, methods currently on the Chess class would remain, but delegate to a single instance of Game when there is only one game read from the PGN file.

JeffML:
I'll go forward on this, breaking it into two commits:

1.  move all game-specific methods and data to a Game object type
    1.  This commit will be fully compatible with the existing Chess.js API
    1.  Current top-level methods related to a game will delegated to a [monostate-ish](https://www.simplethread.com/the-monostate-pattern/) Game instance
2. Assuming the first commit is accepted, the second commit will add arguments for parsing of a multi-game PGN file to the appropriate method(s). 
    1. The intent here is to also maintain backward compatibility with the existing API, namely treating multi-game PGN files as single-game unless otherwise indicated by the client via flags or use of new methods

neofight78:
I think I'm fine with a basic facility for dealing with multi-game PGN files as long as the scope is kept tight. Anything like searches or indexes etc probably should be a different project.

In terms of implementation, currently the Chess class is somewhat of a "God" class. One the one hand it's simple to understand and get up and running with, but on the other hand the mixing of game and board state can make things awkward at times. Mixing in multi game pgn support would probably muddy the water further.

With RAV support coming down the line too it perhaps worth thinking about splitting things out into more focused classes, as long as we don't get too complicated with our abstractions. I want to make sure the library is still quicky and easy to get started with.

[...]

PR #568, opened 2025-10-05
Multigame stage1a

JeffML:
This encapsulates all game-related methods and state to Game.ts. Constants and other shared state moved to types.js. 

This is a backward-compatible refactoring of Chess.js; Chess methods now delegate to a Game instance where warranted.

If this PR is accepted, then I can proceed to Stage 2: multi-game PGN support for chess.js


JeffML:
#561

neofight78:
I've not looked at it in detail yet but these are my initial thoughts:

* types.ts - Bit of a nit pick, but this has constants and functions too. I'd call it common.ts or something more generic. But the idea is good / essential. I've had to do the same when work on my RAV POC.
* Move.ts - Yep, good call to pull this out into it's own file. I'd prefer to keep the filenames all lowercase though (sorry, another nitpick).
* Game.ts - This one I'm not so sure about. Keeping the Chess class as essentially the game object and splitting out board state, and adding something new to represent a PGN feels like a more natural direction to me.

PR #569, opened 2025-10-06
Add NAGs support

brobert1:
### Previous Implementation (Glyph-based)
- Stored annotations as text symbols (e.g., `'±'`, `'∞'`)
- Only one glyph per position
- Used string keys like `$14`, `$16` in mapping
- Limited to predefined glyph symbols

### New Implementation (NAG-based)
- Stores annotations as numeric NAG codes (e.g., `16`, `13`)
- **Supports multiple NAGs per position** (stored as `NAG[]`)
- Uses numeric keys directly (e.g., `14`, `16`)
- Provides utility function to convert NAGs to glyphs

### DIFF
--- a/src/chess.ts
+++ b/src/chess.ts
@@ -95,6 +95,33 @@ export const SUFFIX_LIST = ['!', '?', '!!', '!?', '?!', '??'] as const
 
 export type Suffix = (typeof SUFFIX_LIST)[number]
 
+export const GLYPH_MAP = {
+  '$7': '□', // Only move
+  '$22': '⨀', // Zugzwang
+  '$10': '=', // Equal position
+  '$13': '∞', // Unclear position
+  '$14': '⩲', // White is slightly better
+  '$15': '⩱', // Black is slightly better
+  '$16': '±', // White is better
+  '$17': '∓', // Black is better
+  '$18': '+−', // White is winning
+  '$19': '-+', // Black is winning
+  '$146': 'N', // Novelty
+  '$32': '↑↑', // Development
+  '$36': '↑', // Initiative
+  '$40': '→', // Attack
+  '$132': '⇆', // Counterplay
+  '$138': '⊕', // Time trouble
+  '$44': '=∞', // With compensation
+  '$140': '∆', // With the idea
+} as const
+
+export const GLYPH_LIST = Object.values(GLYPH_MAP)
+
+export type GlyphKey = keyof typeof GLYPH_MAP
+
+export type Glyph = (typeof GLYPH_LIST)[number]
+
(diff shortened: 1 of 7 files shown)


src/chess.ts
```diff
@@ -2248,6 +2285,11 @@ export class Chess {
           if (suffixAnnotation) {
             this._suffixes[this.fen()] = suffixAnnotation as Suffix
           }
+
+          // Store NAGs directly as numbers
+          if (node.nags && node.nags.length > 0) {
+            this._nags[this.fen()] = node.nags.map((nag) => parseInt(nag, 10))
```
neofight78: Probably better to tweak the parser so that nags are already returned as numbers

src/chess.ts
```diff
@@ -2749,6 +2800,68 @@ export class Chess {
     return old
   }
 
+  /**
+   * Get the NAGs for the given position (or current one).
+   */
+  public getNags(fen?: string): NAG[] | undefined {
```
neofight78: Again better without undefined.

src/chess.ts
```diff
@@ -2749,6 +2800,68 @@ export class Chess {
...
+  public setNags(nags: NAG[], fen?: string): void {
+    const key = fen || this.fen()
+    this._nags[key] = [...nags] // Create a copy
+  }
+
+  /**
+   * Remove all NAGs for the given position (or current).
+   * Returns the removed NAGs.
+   */
+  public removeNags(fen?: string): NAG[] | undefined {
```
neofight78: Again better without undefined

src/chess.ts
```diff
@@ -2713,6 +2760,10 @@ export class Chess {
         entry.suffixAnnotation = suffixAnnotation
       }
 
+      if (nags !== undefined && nags.length > 0) {
+        entry.nags = nags
+      }
```
neofight78: I thing if we are keeping this._nags free of empty arrays, we don't need the length check.

src/chess.ts
```diff
@@ -2680,27 +2722,32 @@ export class Chess {
     fen: string
     comment?: string
     suffixAnnotation?: string
+    nags?: NAG[]
```
neofight78: We should probably exclude undefined. Returning an empty array when there are no NAGs is sufficient and simpler for client code to handle.

neofight78 (commented):
A few notes. Basically I think we are better of just returning an empty array for no NAGs rather than using undefined.

brobert1:
Hello, @neofight78!
I have made the adjustments you have requested. Please take a look and let me know if it's ok now.

neofight78:
@brobert1 This looks great, thank you!

Issue #542, opened 2025-06-28
Counting attackers, bearing in mind batteries


neilgd:
Hi,

As per https://github.com/jhlywa/chess.js/issues/186, is there any way of including batteries in the attackers and defenders?

I would like to know if a piece is 'safe', but to do that it's important to take batteries into account.

For example, in this FEN:

rn3rk1/ppp2ppp/8/2bp2q1/4n3/1BN2Q2/PPPP2PP/R1B1KR2 b Q - 4 10

I would like to be able to highlight f7 as 'semi-protected' - i.e. it has the same number of attackers and defenders.

![Image](https://github.com/user-attachments/assets/b26ae9a6-e4be-4e1a-969c-6fcbc8fd982c)

Thanks!

neofight78:
This is not implemented yet, as you are the first to ask for it (I think) . It may be added to a future version if someone writes the code for it.

pabrams:
I'm working on this, if nobody else is.

pabrams:
If anyone needs this functionality, I added an `xray` parameter to the `attackers` method. When `xray=true`, batteries are taken into account. So the number of white attackers of f7 in the above picture would be 2 when `xray=true`, but only 1 when `xray=false`.

However, although the PR was open for several months it was never approved (it seems as though jhlywa/chess.js is no longer actively maintained, but I'm not sure). You can use my fork at [npm: @pabrams/chess.js](https://www.npmjs.com/package/@pabrams/chess.js/v/1.5.4) or [github: pabrams/chess.js](https://github.com/pabrams/chess.js).

JeffML:
Implemented in [chessPGN v1.2.0](https://github.com/JeffML/chessPGN) as an 
optional `{ xray: true }` parameter on `attackers()` and `isAttacked()`. 
Counts pieces attacking through blockers on the same ray — e.g., a rook on a1 
attacks a8 through a queen on a2. Cannot xray through knights or your own king.

```js
chess.attackers('a8', 'w', { xray: true })  // ['a1']
chess.isAttacked('a8', 'w', { xray: true })  // true
```

Issue #186, opened 2018-09-14
How to find defended and attacked pieces?


vitogit:
What´s the best way to get the number of defenders and attackers for each square? 
I want to show in the board what pieces are unguarded and the "influence" for each square, this is the number of attackers - defenders for that square.
Maybe if the legal moves method had an option to include the squares that already have an own piece it could work.

mliebelt:
I would like to have that functionality as well, I will give it a try. I see the following options (with different possible implementations then):

* 2 different calls, one for defended pieces (or squares), one for attacked pieces (or squares).
* each time the option to do that for the turn color, the other color, or both. Here an example
  * Searching for defended pieces, you have to simulate that first an attack is done, and then the defenders are computed. Currently, chess.js does not support defending moves.
  * Searching for attacking pieces, this will normally work in the current position for the color that has the turn. To get attacking pieces for the other color, you have to change the turn of the current position.

So the API could look like this:

* `attacked_pieces(color)`
  * color: default is null ==> denotes the one that has the turn.
  * BLACK | WHITE: defines for which color to search for attacked_pieces
  * BOTH: search for both colors
* `defended_pieces(color)` with the same semantics
* Both return something like `{'bd8': ['ne6', 'rd1'], 'qc7': ['ne6', 'bf4'], 'nf7': []}` ==> Bishop on d8 is attacked by Knight on e6 and by Rook on d1 ...

vitogit:
I did it in a hacky way and used it here https://jsfiddle.net/m5q6fgtb/859/ (When you move a piece it show number of defenders and attackers per square).

To know the numbers of defenders for my piece (Same method can be used to know the attackers), I will remove that piece and put the enemy queen
and then check how many of my pieces can capture it, so that's the  number of defenders. Is not tested much so maybe it's wrong,but it can help someone.

```
function countSquareDefenders(boardElement, fen, square, me) {   
    var chess = new Chess(fenForOtherSide(fen));

    var oppositeColor = chess.turn() == 'w' ? 'b' : 'w'
    var queenSquare = squaresOfPiece(fen, oppositeColor, 'q');
    var somePiece = chess.remove(square); // remove my piece if any
    chess.remove(queenSquare); // remove his queen
    chess.put({ // put queen in the square
        type: 'q',
        color: oppositeColor
    }, square);    
  
    var moves = chess.moves({verbose: true, legal: false})
    var defendersCount = moves.filter(m => m.to == square && m.flags =='c').length
    
[...]

PR #572, opened 2025-10-25
Add battery support to attackers method via xray parameter

pabrams:
- Addresses #542

### DIFF
--- a/__tests__/attackers.test.ts
+++ b/__tests__/attackers.test.ts
@@ -1,10 +1,10 @@
 import { Chess, Color, SQUARES, WHITE, BLACK } from '../src/chess'
 import { expect, test } from 'vitest'
 
-function getAttackerCount(chess: Chess, color: Color) {
+function getAttackerCount(chess: Chess, color: Color, xray: boolean = false) {
   return Array.from(
     { length: 64 },
-    (_, i) => chess.attackers(SQUARES[i], color).length,
+    (_, i) => chess.attackers(SQUARES[i], color, xray).length,
   )
 }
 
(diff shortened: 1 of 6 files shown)


neofight78:
I'm not sure the logic is right. I think in the existing loop, if we meet a blocker, check if it can attack in the same direction, if so carry on with that loop. A second loop seems unnecessary and more complicated to me. Also I think probably some more tests are needed to ensure everything works as expected.

pabrams:
@neofight78 How does this look, now? I may not have interpreted the scope correctly; I based it on comments (e.g. @jhlywa's about an xray parameter) in #186 , but I'm happy to adjust my solution, and/or add more test cases. I wouldn't mind a little help on the testing side. If anyone has some test case suggestions I'll gladly add them.

neofight78:
That's a good step forward, thank you.

I think you've coded it so you can x-ray through a pawn, but not a king. However the king itself can be counted as an attacker. So I'm not sure that's consistent? (See this example: 6k1/8/8/5n2/4K3/3B4/8/8 b - - 0 1)

pabrams:
IIRC, king xrays happen in certain situations, not others. For example, you can't xray through your own king, because it can't be captured? Something like that. You can xray through the opponent's king to the square immediately behind it, to show that that square is attacked, and that prevents the king from moving there. However, king xray behavior might be specific to my use case. But I'll add your test case tomorrow and take another look; even if I'm right, the behavior requires more explaining. It's been a minute since I worked on this.

pabrams:
My code has the following: 
```
    const canXrayThrough = (p: Piece) =>
      !((p.color === attackingPiece.color && p.type === 'k') || p.type === 'n')
```
This means we can never xray through a knight, and we can only xray through the king if it's the opposite color of the attacking piece. I think this makes sense, because in order for a battery to have any effect on the target square, the first pieces in the battery must be capturable. So if the target square is a8, and there's a black king on a7, and a black rook on a6, the black rook's attack has no effect on a8, because the black king cannot be captured (e.g. by a white rook on b8). The black rook on a6 cannot "see" a8 because its own king is in the way. This is because the Black rook on a6 will never be able to get to a8; the black king can capture a piece on a8, conceivably, but the black rook cannot, because its own king would always be there before the rook can get there, and no white piece can remove the black king from a8.

If it was a white king on a7 instead of a black one, then the black rook on a6 would be able to "see" a8; we want its attack to count, so that we know the white king cannot move there, because the the black rook can xray through the white king to the first square behind the white king (but no further).

There might still be some gaps in the logic, but I've been using the functionality in an app that shows board territory control, and I haven't found any problems, yet.

pabrams:
Not sure if this helps or not, but [here's an app I made](https://pabrams.github.io/monkey-drill), which uses the code in this PR. 

I'll be happy to change the xray feature behavior, if someone can make a convincing argument for some other way it should behave.

8/2B5/2k5/8/2R5/8/2r5/7K b - - 0 1
can-xray-thru-own-king-2025-12-16_21-30

2R5/1rB1R3/2k5/2r5/2R5/6b1/2r5/7K b - - 0 1
cant-xray-thru-own-king-2025-12-16_21-27

pabrams:
@neofight78 Is there some different behavior you'd like to see, here? I may be biased by my own use case; it's possible I need to move some of the behavior away from chess.js and into my own app.

neofight78:
TBH When the idea was first floated, I didn't really think about this deeply enough. I only really had basic battery scenarios and attack/defence from different directions in mind.

So I think you can't really x-ray through your own king because it's not possible for a recapture to happen (as you rightly point out). But you can x-ray through opposite colour pieces potentially, but maybe not if there are attacks from different directions? I guess here we are having to think carefully about recapturing sequences?

pabrams:
I was hoping we wouldn't have to get into all that complexity, right now. In the case of xray-thru-king, it seems straightforward, but considering all possible capturing sequences from all possible directions seems hopelessly complicated, to me. There would no longer be a clear answer about which pieces attack a given square, because it would depend on what order the players chose to execute the captures (I think?)

What do you think about possibly tackling that as a separate issue, and just documenting the limitations, for now? I could add language to the effect that attacks are counted even if those attacks are meaningless.... The only reason we don't xray through our own king is because it will _never_ be possible for the attacking piece to reach the target square, as long as the king is in the way on the same ray, no matter what the rest of the board looks like. But in the case of four black rooks stacked up along the same ray, it's conceivable that there could be white pieces elsewhere on the board that render all the black rooks' attacks meaningful (and therefore, we count them all). 

tl;dr: When we calculate x-ray attackers, we only consider each ray, from each potential attacking piece to the target square. We don't attempt to add up all the white and black attacks (or execute all possible captures from other directions, because that's indeterminate, anyway) to ensure that each attack "matters".

pabrams:
Closing this for lack of interest - if anyone wants the x-ray functionality they can use my fork [npm: @pabrams/chess.js](https://www.npmjs.com/package/@pabrams/chess.js/v/1.5.2)

PR #578, opened 2026-01-17
Fix: Pawns on edge ranks should have no chess.moves()

### LINKED ISSUE #577: BigInt error when generating moves for pawns on edge ranks
Error happens in cases like:

```
chess.put({ type: 'p', color: 'w' }, 'h8')
chess.moves({ square: 'h8', verbose: true }))
```

Output:
```
TypeError: Cannot mix BigInt and other types, use explicit conversions
    at Chess._movePiece (/app/node_modules/src/chess.ts:1737:24)
    at Chess._makeMove (/app/node_modules/src/chess.ts:1764:10)
    at Chess._moveToSan (/app/node_modules/src/chess.ts:2279:10)
    at new Move (/app/node_modules/src/chess.ts:160:35)
    at /app/node_modules/src/chess.ts:1452:34
    at Array.map (<anonymous>)
    at Chess.moves (/app/node_modules/src/chess.ts:1452:20)
```

some1else:
Closes #577 

Pawn on H8 throws a BigInt error when asked about `chess.moves({ square: 'h8' })`

```
TypeError: Cannot mix BigInt and other types, use explicit conversions
    at Chess._movePiece (/app/node_modules/src/chess.ts:1737:24)
    at Chess._makeMove (/app/node_modules/src/chess.ts:1764:10)
    at Chess._moveToSan (/app/node_modules/src/chess.ts:2279:10)
    at new Move (/app/node_modules/src/chess.ts:160:35)
    at /app/node_modules/src/chess.ts:1452:34
    at Array.map (<anonymous>)
    at Chess.moves (/app/node_modules/src/chess.ts:1452:20)
```

### DIFF
--- a/__tests__/regression.test.ts
+++ b/__tests__/regression.test.ts
@@ -29,6 +29,15 @@ describe('Regression Tests', () => {
     expect(chess.moves().join(' ')).toBe('Kd2 Ke2 Kxf2 Kf1 Kd1')
   })
 
+  it('Github Issue ##577 - pawn on edge rank should have no moves', () => {
+    const chess = new Chess()
+    chess.clear()
+    chess.put({ type: 'k', color: 'w' }, 'e1')
+    chess.put({ type: 'k', color: 'b' }, 'e8')
+    chess.put({ type: 'p', color: 'w' }, 'h8')
+    expect(chess.moves({ square: 'h8', verbose: true })).toEqual([])
+  })
+
   it('Github Issue #85 (white) - SetUp and FEN should be accepted in loadPgn', () => {
     const chess = new Chess()
     const pgn = [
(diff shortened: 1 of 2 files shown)


neofight78:
I'm not sure whether to merge this yet, as a pawn on h8 is not a legal position. It  maybe better to tighten up / change the way the board can be set up.

some1else:
Depends on how strictly you'd like to lock the library down to modern chess rules. There are fringe regimes such as the British Chess Association "dead pawn" rule, and the Burmese Chess promotion rules, under which a pawn can legally occupy the promotion square: https://en.wikipedia.org/wiki/Promotion_%28chess%29

You'll want to reconcile the chosen policy with `{ skipValidation: true }` options for the constructor, as that is likely another code path through which it's possible to place a pawn there.

I'm prototyping a rules-free software chess set mimicking a physical board & pieces, where it's in line with the concept to allow such positions: https://shah-baby.netlify.app. It'd be nice if I could still use chess.js for state management.

PR #583, opened 2026-08-08
add isCheck() method to Move class

MohamedSayed0573:
Adds an `.isCheck()` helper method to the `Move` class that returns `true` if the move puts the opponent's king in check or checkmate.

  While `chess.isCheck()` returns the check status of the active board position, `move.isCheck()` is useful when inspecting move objects returned by:
    - `chess.moves({ verbose: true })` — to identify candidate moves that deliver check without manually parsing `san`.
    - `chess.history({ verbose: true })` — to check historical move properties without modifying or querying the active board state.

   ## Changes

    - Implemented `isCheck()` on `Move` in `src/chess.ts`.
    - Updated unit test assertions across `__tests__/move.test.ts`.
    - Updated documentation in `website/docs/index.md`.
    - Updated public API signatures in `etc/chess.js.api.md`.


MohamedSayed0573:
@greptileai

PR #585, opened 2026-08-26
Fix put/remove always clearing the en passant square

seojoonrp:
## Problem

`_updateEnPassantSquare()` re-checks whether an en passant square is still
reachable after the board is edited. Its first condition is:

```ts
this._board[startSquare] !== null ||
this._board[this._epSquare] !== null ||
```

`_board` is a sparse `new Array<Piece>(128)`, so empty squares read back as
`undefined`, not `null` — `_clear()` removes entries with `delete`, and no code
path ever assigns `null`. `undefined !== null` is always true, so the condition
short circuits and the function unconditionally clears the en passant square.

Any `put()` or `remove()` — including on a square with nothing to do with the
en passant capture — silently discards a legal move:

```ts
const chess = new Chess('rnbqkbnr/ppp1p1pp/8/3pPp2/8/8/PPPP1PPP/RNBQKBNR w KQkq f6 0 3')
chess.moves().includes('exf6')                   // true
chess.put({ type: KNIGHT, color: WHITE }, 'h3')  // unrelated square
chess.moves().includes('exf6')                   // false
chess.fen()                                      // "... w KQkq - 0 3", f6 is gone
```

`load()` goes through the private `_put()` and never calls this function, so
loading a FEN or PGN is unaffected. Only `put()` and `remove()` are.

## Why it went unnoticed

Coverage shows the rest of the function has never executed since #398
introduced it — v8 reports the remaining conditions and the whole
"is a pawn still able to capture?" block as uncovered.

The 16 assertions added alongside it in `put.test.ts` and `remove.test.ts` are
all `expect(chess.moves()).not.toContain(...)`. A function that always clears
the en passant square passes every one of them.

## Changes

1. Use loose equality in the two board checks, so the remainder of the function
   becomes reachable.
2. Clear `_fenEpSquare` alongside `_epSquare` in the branch where the board no
   longer describes a two square pawn move. `_fenEpSquare` was introduced in
   #546 but not threaded through this function, so
   `fen({ forceEnpassantSquare: true })` kept reporting a square that is no
   longer on the board. It is deliberately retained in the second branch, where
   the pawn move is intact but no capturing pawn is left — matching what
   `_makeMove()` does after a big pawn move with no adjacent enemy pawn.
3. Add tests for the preserving direction, and for both `_fenEpSquare` outcomes.

Happy to split (2) into its own PR if you would rather keep this to the one
line fix.

## Verification

- `npm run check` passes: 472 tests (464 existing + 8 new), build and
  api-extractor clean. No public API change — the fix is in a private method.
- 7 of the 8 new tests fail without the source change. The eighth pins the
  previously unreachable "no capturing pawn left" branch, which behaves the
  same either way.


JeffML:
For anyone tracking this: chessPGN (a TypeScript fork of chess.js maintaining API parity) had the identical bug in its put()/remove() en passant handling — the sparse board array returns undefined for empty squares, but the guard checked !== null, so it was always true and the en passant square was cleared on any board edit, not just relevant ones.

Fixed in v1.4.2 by switching to loose equality (!=) on the emptiness checks, mirroring this PR's approach. Thanks for the clear diagnosis here — it made porting the fix straightforward.

seojoonrp:
Thanks for confirming — good to know the diagnosis ported cleanly.

This one is still waiting on workflow approval, so CI hasn't run against it yet. If you have a moment to take a look, or to flag it for someone who can approve the run, I'd appreciate it.

